PR #129, opened 2020-11-20
Add container.id to resource attributes

### LINKED ISSUE #126: Add container ID to span attributes
A similar to https://github.com/Traceableai/javaagent/blob/0baa5ef6282b3ece0d80fb902429b236b18d7a75/modules/core/src/main/java/ai/traceable/agent/modules/core/AgentEnvironmentServiceImpl.java#L13

pavolloffay:
Resolves #126

### DIFF
--- a/smoke-tests/src/test/java/org/hypertrace/agent/smoketest/SpringBootSmokeTest.java
+++ b/smoke-tests/src/test/java/org/hypertrace/agent/smoketest/SpringBootSmokeTest.java
@@ -113,6 +113,10 @@ public class SpringBootSmokeTest extends AbstractSmokeTest {
                 .map(attribute -> attribute.getValue().getStringValue())
                 .count()
             > 0);
+    // TODO test the container ID once we switch from Zipkin exporter
+    // Zipkin exporter does not add resource attributes to span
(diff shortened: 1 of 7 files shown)


smoke-tests/src/test/java/org/hypertrace/agent/smoketest/SpringBootSmokeTest.java
```diff
@@ -113,6 +113,10 @@ public void get() throws IOException {
                 .map(attribute -> attribute.getValue().getStringValue())
                 .count()
             > 0);
+    // TODO test the container ID once we switch from Zipkin exporter
+    // Zipkin exporter does not add resource attributes to span
+    // https://github.com/open-telemetry/opentelemetry-java/issues/1970
```
pavolloffay, line 118: cc) @davexroth https://github.com/open-telemetry/opentelemetry-java/issues/1970

davexroth: hmm, so what do we want to do in the mean time, as the platform needs the container id?  Should we just add it as an attribute for now?

PR #138, opened 2020-11-24
Dynamically disable wrapping of servlet types if cast exception happens

pavolloffay:

### DIFF
--- a/instrumentation/servlet/servlet-3.1/src/main/java/io/opentelemetry/instrumentation/hypertrace/servlet/v3_1/Servlet31Advice.java
+++ b/instrumentation/servlet/servlet-3.1/src/main/java/io/opentelemetry/instrumentation/hypertrace/servlet/v3_1/Servlet31Advice.java
@@ -42,7 +42,7 @@ public class Servlet31Advice {
   public static boolean start(
       @Advice.Argument(value = 0, readOnly = false) ServletRequest request,
       @Advice.Argument(value = 1, readOnly = false) ServletResponse response,
-      @Advice.Local("rootStart") Boolean rootStart) {
+      @Advice.Local("rootStart") boolean rootStart) {
 
     if (!HypertraceConfig.isInstrumentationEnabled(
         Servlet31InstrumentationName.PRIMARY, Servlet31InstrumentationName.OTHER)) {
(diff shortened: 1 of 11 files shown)


instrumentation/servlet/servlet-3.1/src/main/java/io/opentelemetry/instrumentation/hypertrace/servlet/v3_1/Servlet31Advice.java
```diff
@@ -97,7 +97,14 @@ public static Object start(
   public static void stopSpan(
       @Advice.Argument(0) ServletRequest request,
       @Advice.Argument(1) ServletResponse response,
+      @Advice.Thrown Throwable throwable,
       @Advice.Local("rootStart") Boolean rootStart) {
+
+    if (throwable instanceof ClassCastException) {
+      ClassCastException cast = (ClassCastException) throwable;
+      System.out.println(cast.getMessage().contains("hypertrace"));
```
jcchavezs, line 105: Worth to mention why this piece is important? Maybe abstracting `"hypertrace"` into something meaningful like `HYPERTRACE_OWN_PREFIX` or something that makes sense?

pavolloffay: ignore this code completely this is just a nasty example.

The right solution could parse the class name and store it into the global config. The global config would be used by instrumentation points that create wrappers. If no exception is recorded for the wrapped type the wrapping would be applied.

pavolloffay: Another interesting fact in this example is the place where the exceptions are caught. In servlet this can be done in exit filter advice. In other frameworks it can be a different place.

The problem also is if the casting would be inside user defined try-catch(Throwable) block, this instrumentation would not catch this. I have tried to define adice for `ClassCastException` but that does not seem to be triggered.

instrumentation/servlet/servlet-3.1/src/test/java/io/opentelemetry/instrumentation/hypertrace/servlet/v3_1/Servlet31InstrumentationTest.java
```diff
@@ -79,6 +91,58 @@ public void postJson() throws Exception {
         spanData
             .getAttributes()
             .get(HypertraceSemanticAttributes.httpResponseHeader(TestServlet.RESPONSE_HEADER)));
-    server.stop();
+  }
+
+  @Test
+  public void requestCast() throws Exception {
```
pavolloffay, line 97: @davexroth @surajpuvvada here is an example of the cast exception. The first request fails, however the second passes but the attributes that are captured via wrappers are omitted. This is the best what we can do.

PR #140, opened 2020-11-26
Apache httpclient wrapping free body instrumentation

pavolloffay:
There are two instrumentations in this PR

### 1. add result of `InputStream.read` only when the client reads it

Done in `ApacheClientInstrumentationModule` and `InputStreamInstrumentaionModule`

This instrumentation instruments all `InputStream` read methods. The returned `byte[]` or `byte` are added to our buffer stored in a globally accessible registry/map. The final body value is added to a span when the read method returns `-1` (no more content).

The instrumentation advice is applied to all read methods, but we only record reads for input streams that are returned from instrumented objects e.g. `org.apache.http.HttpEntity#getContent()` (when this is called the input stream is put into global registry/map which is then accessed from the `InputStream` instrumentation). 

#### Issues
1. the client span is already finished when the full body is assembled - this can be solved by creating a follow up span with only the body content and the client span is set as a parent. We will need a custom logic in the backend to merge these spans.
2. if we use this approach for server input stream the request will be blocked just before the body is fully assembled. If the app does not read the body the blocking will not happen. Using this for servlet blocking needs more thoughts bc the instrumentation would get the body outside of `doFilter`/`service` methods so to block the request it would have to for example throw an exception.  

### 2. read full content of `InputStream` before the response object is returned

The client instrumentation reads full content of response `InputStream` and adds response body object to the client span.
The `InputStream` instrumentation now creates `ByteArrayInputStream` from the response body and all methods of the original input stream are instrumented - the original method body is skipped (like we do for blocking) and the instrumentation instead delegates calls to the newly created byte stream.

#### Issues
1. The newly created byte stream is not exposed to the application via type but the behavior of its method might differ.
2. more complex instrumentation than 1. as we have to instrument all input stream methods.
3.  I still could not make it work for `DynamoDB-SpringBoot-POC` - the instrumentation read the inputstream very early. I think it's failing bc the stream is not ready yet. 

TODOs

- [x] use the correct encodings
- [x] use concurrent maps
- [x] shade weak concurrent map
- [x] add data capture configuration
- [x] call toString(charset) on the byte output stream instead of getting the array and converting to stream separately
- [ ] could we use a more specific type matcher for the input/output streams? --- leaving this open for now, I will perf test this and see how it behaves on the widest matcher.


pavolloffay:
Output from the `DynamoDBSpringBootPOC`

```bash
2020-11-30 14:55:10.968  INFO 30200 --- [           main] c.c.poc.DynamoDBSpringBootPOC.App        : List object: {"id":"e21b33aa-6c4d-4005-bbbe-d6e085d95081","serviceName":"AWS DynamoDB","serviceHomePageUrl":"https://aws.amazon.com/dynamodb/?nc2\u003dh_m1"}
2020-11-30 14:55:10.968  INFO 30200 --- [           main] c.c.poc.DynamoDBSpringBootPOC.App        : List object: {"id":"b36b4ef1-6090-41b7-8379-34f09c0d89c1","serviceName":"AWS DynamoDB","serviceHomePageUrl":"https://aws.amazon.com/dynamodb/?nc2\u003dh_m1"}
2020-11-30 14:55:10.968  INFO 30200 --- [           main] c.c.poc.DynamoDBSpringBootPOC.App        : List object: {"id":"997fadb0-bb5c-41c3-8c16-93d296119fcf","serviceName":"AWS DynamoDB","serviceHomePageUrl":"https://aws.amazon.com/dynamodb/?nc2\u003dh_m1"}
2020-11-30 14:55:10.969  INFO 30200 --- [           main] c.c.poc.DynamoDBSpringBootPOC.App        : List object: {"id":"bab6325f-93fc-46da-b192-7032351bd821","serviceName":"AWS DynamoDB","serviceHomePageUrl":"https://aws.amazon.com/dynamodb/?nc2\u003dh_m1"}
2020-11-30 14:55:10.969  INFO 30200 --- [           main] c.c.poc.DynamoDBSpringBootPOC.App        : List object: {"id":"9b995e00-e942-43f4-a65a-b6cbe527436e","serviceName":"AWS DynamoDB","serviceHomePageUrl":"https://aws.amazon.com/dynamodb/?nc2\u003dh_m1"}
captured http.request.body readable body is {"TableName":"awsServices"}
Captured http.response.body attribute: {"Items":[{"serviceName":{"S":"AWS DynamoDB"},"serviceHomePageUrl":{"S":"https://aws.amazon.com/dynamodb/?nc2=h_m1"},"id":{"S":"e21b33aa-6c4d-4005-bbbe-d6e085d95081"}},{"serviceName":{"S":"AWS DynamoDB"},"serviceHomePageUrl":{"S":"https://aws.amazon.com/dynamodb/?nc2=h_m1"},"id":{"S":"b36b4ef1-6090-41b7-8379-34f09c0d89c1"}},{"serviceName":{"S":"AWS DynamoDB"},"serviceHomePageUrl":{"S":"https://aws.amazon.com/dynamodb/?nc2=h_m1"},"id":{"S":"997fadb0-bb5c-41c3-8c16-93d296119fcf"}},{"serviceName":{"S":"AWS DynamoDB"},"serviceHomePageUrl":{"S":"https://aws.amazon.com/dynamodb/?nc2=h_m1"},"id":{"S":"bab6325f-93fc-46da-b192-7032351bd821"}},{"serviceName":{"S":"AWS DynamoDB"},"serviceHomePageUrl":{"S":"https://aws.amazon.com/dynamodb/?nc2=h_m1"},"id":{"S":"9b995e00-e942-43f4-a65a-b6cbe527436e"}}],"Count":5,"ScannedCount":5}
```

instrumentation/apache-httpclient-4.0/src/main/java/io/opentelemetry/instrumentation/hypertrace/apachehttpclient/v4_0/ApacheHttpClientUtils.java
```diff
@@ -0,0 +1,109 @@
...
+          currentSpan, HypertraceSemanticAttributes.HTTP_REQUEST_BODY.getKey(), entity);
+    }
+  }
+
+  public static void traceEntity(Span span, String bodyAttributeKey, HttpEntity entity) {
+    if (entity == null) {
+      return;
+    }
+
+    if (entity.isRepeatable()) {
```
surajpuvvada, line 82: we won't capture if it isn't repeatable ?

pavolloffay: see line 107

pavolloffay: It captures bodies for repeatable but also non-repeatable streams. There is also a unit test to verify it.

surajpuvvada (approved):
This was a quick one :)

pavolloffay:
thanks @surajpuvvada, I will perf test tomorrow, really looking forward how it will play out. If it works well we could use the same approach for other frameworks even servlet.

Issue #65, opened 2020-10-29
Verify that HT agent can run alongside DataDog agent and AppDynamics


pavolloffay:
* check the memory consumption of the agent
* run agents in all orders HT, DD and DD and HT

PR #146, opened 2020-12-01
Ignore classes from other agents

pavolloffay:
Related to #65 

Other agents classes are already excluded in OTEL upstream. The https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1812 and https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1813 adds exclusion for datadog and newrelic to the upstream. Our fix might be needed since some classes used by datadog are not shaded and live in agent's classloader (e.g. okhttp exporter).


pavolloffay:
Apparently, this is causing issues. In some cases, it seems to break instrumentations. The agent customizer is called right before `installOn` and after some ignores are added to the build and after instrumentations are loaded.

pavolloffay:
Done in upstream

Issue #109, opened 2020-11-13
Capture HTTP2 Headers (scheme, authority, path, method) gRPC client/server


pavolloffay:
This will have to be done at the grpc-netty level. 

If we could do it in such a way that it works with single instrumentation on `grpc-okhttp` `grpc-netty-shaded` it would be great.

pavolloffay:
#158 added support for capturing client and server HTTP2 headers. However the client instrumentation is not bullet proof - the headers are not captured for the first request if `DelayedStream` is used - maybe they won't be also captured if the channel status switches from READY to other states and back to ready.


```java
/**
 * The current span in Utils.convertClientHeaders is DefaultSpan for the first request. The first
 * request uses io.grpc.internal.DelayedClientTransport and it is called from
 * io.grpc.stub.ClientCalls.blockingUnaryCall. Subsequent requests use
 * io.grpc.stub.ClientCalls.futureUnaryCall - see (1). The DelayedClientTransport calls network in a
 * separate branch after ClientCallImpl or gRPC interceptors. To propagate span to
 * Utils.convertClientHeaders it would have to be started in
 * io.grpc.stub.ClientCalls.blockingUnaryCall.
 *
 * <p>Span is not recording (Default) java.lang.Exception: Stack trace at
 * java.lang.Thread.dumpStack(Thread.java:1336) at
 * io.grpc.netty.Utils.convertClientHeaders(Utils.java:107) at
 * io.grpc.netty.NettyClientStream$Sink.writeHeaders(NettyClientStream.java:124) at
 * io.grpc.internal.AbstractClientStream.start(AbstractClientStream.java:132) at
 * io.grpc.internal.DelayedStream$4.run(DelayedStream.java:197) at
 * io.grpc.internal.DelayedStream.drainPendingCalls(DelayedStream.java:132) at
 * io.grpc.internal.DelayedStream.setStream(DelayedStream.java:101) at
 * io.grpc.internal.DelayedClientTransport$PendingStream.createRealStream(DelayedClientTransport.java:351)
 * at
 * io.grpc.internal.DelayedClientTransport$PendingStream.access$200(DelayedClientTransport.java:334)
 * at io.grpc.internal.DelayedClientTransport$5.run(DelayedClientTransport.java:293) at
 * io.grpc.stub.ClientCalls$ThreadlessExecutor.waitAndDrain(ClientCalls.java:575) (1) at
 *
[...]

PR #158, opened 2020-12-08
Capture HTTP2 headers in netty gRPC client/server

pavolloffay:

Updates #109

### DIFF
--- a/instrumentation/grpc-1.5/src/main/java/io/opentelemetry/instrumentation/hypertrace/grpc/v1_5/Http2HeadersInstrumentationModule.java
+++ b/instrumentation/grpc-1.5/src/main/java/io/opentelemetry/instrumentation/hypertrace/grpc/v1_5/Http2HeadersInstrumentationModule.java
@@ -39,9 +39,9 @@ import net.bytebuddy.matcher.ElementMatcher;
 import org.hypertrace.agent.core.HypertraceSemanticAttributes;
 
 @AutoService(InstrumentationModule.class)
-public class ClientHTTP2HeadersInstrumentationModule extends InstrumentationModule {
+public class Http2HeadersInstrumentationModule extends InstrumentationModule {
 
-  public ClientHTTP2HeadersInstrumentationModule() {
+  public Http2HeadersInstrumentationModule() {
     super(GrpcInstrumentationName.PRIMARY, GrpcInstrumentationName.OTHER);
   }
 
@@ -50,8 +50,15 @@ public class ClientHTTP2HeadersInstrumentationModule extends InstrumentationModu
     return Arrays.asList(new NettyUtilsInstrumentation());
   }
 
+  /**
+   * The server side HTTP2 headers are added in tracing gRPC interceptor. The headers are added to
+   * metadata in {@link GrpcUtils_convertHeaders_Advice}.
+   *
+   * <p>The client side HTTP2 headers are added directly to span in {@link
+   * Utils_convertClientHeaders_Advice}. TODO However it does not work for the first request
+   * https://github.com/hypertrace/javaagent/issues/109#issuecomment-740918018.
+   */
   class NettyUtilsInstrumentation implements TypeInstrumentation {
-
     @Override
     public ElementMatcher<? super TypeDescription> typeMatcher() {
       return failSafe(named("io.grpc.netty.Utils"));
--- a/instrumentation/grpc-1.5/build.gradle.kts
+++ b/instrumentation/grpc-1.5/build.gradle.kts
@@ -72,6 +72,4 @@ dependencies {
     testImplementation("io.grpc:grpc-protobuf:1.5.0")
     testImplementation("io.grpc:grpc-stub:1.5.0")
     testImplementation("io.grpc:grpc-netty:1.5.0")
-    // TODO remove if not needed
-    testImplementation("io.opentelemetry.javaagent.instrumentation:opentelemetry-javaagent-netty-4.1:${versions["opentelemetry_java_agent"]}")
 }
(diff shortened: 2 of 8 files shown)


instrumentation/grpc-1.5/src/main/java/io/opentelemetry/instrumentation/hypertrace/grpc/v1_5/Http2HeadersInstrumentationModule.java
```diff
@@ -0,0 +1,133 @@
...
+          isMethod().and(named("convertClientHeaders")).and(takesArguments(6)),
+          Utils_convertClientHeaders_Advice.class.getName());
+      transformers.put(
+          isMethod().and(named("convertHeaders")).and(takesArguments(1)),
+          GrpcUtils_convertHeaders_Advice.class.getName());
+      return transformers;
+    }
+  }
+
+  static class Utils_convertClientHeaders_Advice {
```
singhalprerana, line 80: nit: GrpcUtils maybe?

instrumentation/grpc-1.5/src/main/java/io/opentelemetry/instrumentation/hypertrace/grpc/v1_5/Http2HeadersInstrumentationModule.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+      Map<ElementMatcher<? super MethodDescription>, String> transformers = new HashMap<>();
+      transformers.put(
+          isMethod().and(named("convertClientHeaders")).and(takesArguments(6)),
+          Utils_convertClientHeaders_Advice.class.getName());
+      transformers.put(
+          isMethod().and(named("convertHeaders")).and(takesArguments(1)),
```
singhalprerana, line 74: I think this is called multiple times from different flows in code.. How are we keeping track when to use and when not to use?

instrumentation/grpc-1.5/src/main/java/io/opentelemetry/instrumentation/hypertrace/grpc/v1_5/Http2HeadersInstrumentationModule.java
```diff
@@ -0,0 +1,133 @@
...
+
+  static class Utils_convertClientHeaders_Advice {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void exit(
+        @Advice.Argument(1) Object scheme,
+        @Advice.Argument(2) Object defaultPath,
+        @Advice.Argument(3) Object authority,
+        @Advice.Argument(4) Object method) {
+
+      Span currentSpan = Java8BytecodeBridge.currentSpan();
```
singhalprerana, line 88: This method can be called when a span does not exist right? We need to add these attributes only when some span has been started.. Are we tracking that? If yes, then span can be null -- need to handle that..

instrumentation/grpc-1.5/src/main/java/io/opentelemetry/instrumentation/hypertrace/grpc/v1_5/Http2HeadersInstrumentationModule.java
```diff
@@ -0,0 +1,133 @@
...
+
+  static class GrpcUtils_convertHeaders_Advice {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void exit(
+        @Advice.Argument(0) Http2Headers http2Headers,
+        @Advice.Return(readOnly = false) Metadata metadata) {
+
+      if (http2Headers.authority() != null) {
+        metadata.put(
+            GrpcSemanticAttributes.AUTHORITY_METADATA_KEY, http2Headers.authority().toString());
```
singhalprerana, line 120: we are changing original user metadata object -- need to be very careful if this does not break anything -- especially since this method is called from different flows in code..

instrumentation/grpc-1.5/src/main/java/io/opentelemetry/instrumentation/hypertrace/grpc/v1_5/Http2HeadersInstrumentationModule.java
pavolloffay: span is never null. It is either `DefaultSpan` and recording events span.

instrumentation/grpc-1.5/src/main/java/io/opentelemetry/instrumentation/hypertrace/grpc/v1_5/Http2HeadersInstrumentationModule.java
pavolloffay: The target class is called `Utils` therefore the `Utils` name.

instrumentation/grpc-1.5/src/main/java/io/opentelemetry/instrumentation/hypertrace/grpc/v1_5/Http2HeadersInstrumentationModule.java
pavolloffay: correct we do not keep track of objects where should be added. I want to avoid using TLS and global map to do this as it is done in the old impl. 

The HTTP2 headers are added under different keys than the gRPC impl uses e.g. `:method.ht` to do not collide with user app. The implementation does not change any objects it just adds more data to metadata object passed around I think this is acceptable also the added headers are not added to the wire.

For completion: this advice does not change client response headers since those headers do not have the HTTP2 we are using (see the comment in the code). If the server side will cause any issues we could remove added headers in the interceptor before passing it to user define interceptors. The removal seems expensive so I would rather skip it for now.

PR #180, opened 2020-12-17
Skip loading the agent if traceable agent has been loaded

pavolloffay:

This is needed for dynamic attach, our attacher might try to attach new agent even though the old one has already been attached.


javaagent/src/main/java/org/hypertrace/agent/instrument/HypertraceAgent.java
```diff
@@ -103,4 +108,13 @@ static String toOtelPropagators(List<PropagationFormat> propagationFormats) {
     }
     return argsMap;
   }
+
+  private static boolean isTraceableAgentLoaded() {
+    try {
+      Class.forName("ai.traceable.agent.bootstrap.AgentContext", true, null);
```
pavolloffay, line 114: this uses bootstrap classloader and this class seems to be there from the beginning. cc) @surajpuvvada

PR #193, opened 2020-12-22
Add netty body, headers capture and filter evaluation

pavolloffay:
This PR adds support for server side netty payload and headers capture. It also adds tests for vertx-3.0 that is based on netty.

TODOs:

- [x] data capture config in netty 4.1
- [x] check`CallDepthThreadLocalMap` in  `ChannelPipelineAddAdvice` in netty-4.1 
- [x] check if charset is set correctly in other instrumenations that were changed in this PR
- [x] add blocking in request handler


instrumentation/apache-httpclient-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/apachehttpclient/v4_0/HttpEntityInstrumentation.java
```diff
@@ -114,8 +114,13 @@ public static void enter(
       if (contentSize <= 0 || contentSize == Long.MAX_VALUE) {
         contentSize = ContentLengthUtils.DEFAULT;
       }
-      ByteArrayOutputStream byteArrayOutputStream =
-          BoundedByteArrayOutputStreamFactory.create((int) contentSize);
+
+      // TODO set charset
+      String encoding =
+          thizz.getContentEncoding() != null ? thizz.getContentEncoding().getValue() : "";
```
surajpuvvada, line 120: what encoding are you planning to use for the case where it is null ?

pavolloffay: this is wrong, the encoding here is not charset. The default charset is `ISO-8859-1`

pavolloffay: It has been fixed in the next commit

PR #225, opened 2021-01-11
Bump to the newest HT config with snake case

pavolloffay:

It seems that our deserialization libraries support snake and camel case at the same time.


pavolloffay:
cc) @davexroth ^^^

PR #250, opened 2021-02-02
lowercase all http header and rpc metadata keys

davexroth:
For consistent lookup of http headers and rpc metadata in the span attribute map, we should always lower case them.

This allows collector processors to lookup headers without having to pre-process the span.


pavolloffay (approved):
Could you please describe the use-case in the PR description?

PR #260, opened 2021-02-10
Bump OTEL to 0.15.0

pavolloffay:

### DIFF
--- a/instrumentation/build.gradle.kts
+++ b/instrumentation/build.gradle.kts
@@ -13,9 +13,7 @@ subprojects {
         implementation("net.bytebuddy:byte-buddy:${versions["byte_buddy"]}")
 
         implementation("io.opentelemetry:opentelemetry-api:${versions["opentelemetry"]}")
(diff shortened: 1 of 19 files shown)


instrumentation/build.gradle.kts
```diff
@@ -13,7 +13,9 @@ subprojects {
         implementation("net.bytebuddy:byte-buddy:${versions["byte_buddy"]}")
 
         implementation("io.opentelemetry:opentelemetry-api:${versions["opentelemetry"]}")
-        implementation("io.opentelemetry.javaagent:opentelemetry-javaagent-tooling:${versions["opentelemetry_java_agent"]}")
+        implementation("io.opentelemetry.javaagent:opentelemetry-javaagent-tooling:${versions["opentelemetry_java_agent"]}") {
+            exclude(group = "io.netty")
```
pavolloffay, line 17: it seems that we cannot use exclude. It breaks uber jar functionality -  the gRPC exporters will not work

PR #267, opened 2021-02-17
Catch exceptions in filter

shashank11p:
Add try catch for loading filter jars and its implementations


filter-api/src/main/java/org/hypertrace/agent/filter/MultiFilter.java
```diff
@@ -33,8 +37,16 @@ public MultiFilter(List<Filter> filters) {
...
     for (Filter filter : filters) {
-      if (filter.evaluateRequestHeaders(span, headers)) {
-        shouldBlock = true;
+      try {
+        if (filter.evaluateRequestHeaders(span, headers)) {
+          shouldBlock = true;
+        }
+      } catch (Throwable t) {
+        logger.warn(
+            String.format(
```
pavolloffay, line 46: no need to use the string format, the logger can format it

pavolloffay (approved):

Issue #217, opened 2021-01-11
Add more smoke test and framework how to add more tests


pavolloffay:
Add smoke tests for gRPC and different servlet containers (similar to https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/master/smoke-tests/matrix).

The images for these should be in control of this project, build as part of the smoke-test job. It should be easy to add smoke tests for new frameworks.

pavolloffay:
We can certainly adopt OTEL smoke tests, my requirements are: 
* test capturing of payloads
* test capturing of headers
* the test images should be built and published to HT docker registry - e.g. ghcr or dockerhub

samarth-gupta-traceable:
@pavolloffay  here the  2 steps I propose to integrate and modify OTEL smoke test with HT java agent:

1. Import current test from OTEL to HT repo and make sure all test are passing with `hypertrace-agent-all.jar`  . This will  require changes in smoke test as well as gradle files. This will include changes for publishing images to HT docker registry. 
2. Enhance test app to send POST requests with required headers & add validation by enhancing test.

PR #268, opened 2021-02-17
Add OTEL smoke test

samarth-gupta-traceable:
Updates #217 

## Description

### Testing

ran ./gradlew build locally 

### Documentation


Makefile
```diff
@@ -7,11 +7,11 @@ assemble:
 
 .PHONY: build
 build:
-	./gradlew build -x :smoke-tests:test --stacktrace
+	./gradlew build -x :smoke-tests:test --debug --stacktrace
```
pavolloffay, line 10: is this needed? I don't like to run debug locally. As below in the `smoke-test` target

gradle/dependencies.gradle
```diff
@@ -0,0 +1,97 @@
+def groovyVer = "2.5.11"
```
pavolloffay, line 1: Do we need this file and the `java.gradle`? 

in OTEL repo these are reused for other modules. since we have only smoke tests module we could copy there only what is needed. For instance we define versions in our root build script.

pavolloffay (commented):
Looks good, thanks for the PR 👍 

We should try to remove those two gradle scripts and reduce CI run time

Makefile
samarth-gupta-traceable: have removed the debug option

gradle/dependencies.gradle
samarth-gupta-traceable: @pavolloffay  have removed dependencies.gradle but not java.gradle.

pavolloffay (approved):

pavolloffay:
@samarth-gupta-traceable the next step will be to modify the apps to include `/echo` endpoint and add a test with posts JSON (and receives it).

PR #281, opened 2021-03-03
Publish jars to maven central repository

ravisingal:
snapshot builds are available at: https://s01.oss.sonatype.org/content/repositories/snapshots/org/hypertrace/agent/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,6 +72,7 @@ jobs:
       - docker_login
       - run: echo "Releasing version" && ./gradlew printVersion
       - run: ./gradlew publish
+      - run: ./gradlew uploadArchives
       - run: DOCKER_TAG=${CIRCLE_TAG} make docker docker-push
       - run:
           name: "Copy artifacts to workspace"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,12 +100,10 @@ workflows:
   build-and-publish:
     jobs:
       - build:
-          context: hypertrace-publishing
           filters:
             tags:
               only: /.*/
       - muzzle:
-          context: hypertrace-publishing
           filters:
             tags:
               only: /.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,10 +100,12 @@ workflows:
   build-and-publish:
     jobs:
       - build:
+          context: hypertrace-publishing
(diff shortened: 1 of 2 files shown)


pavolloffay:
@ravisingal the build failed

.circleci/config.yml
```diff
@@ -72,6 +72,7 @@ jobs:
       - docker_login
       - run: echo "Releasing version" && ./gradlew printVersion
       - run: ./gradlew publish
+      - run: ./gradlew uploadArchives
```
aaron-steinfeld, line 75: I thought we had migrated everything off circleci? I guess this repo was missed - would it make sense to migrate it before adding more circleci workflows?

build.gradle.kts
```diff
@@ -69,5 +71,39 @@ subprojects {
                 junitXml.isOutputPerTestCase = true
             }
         }
+        "uploadArchives"(Upload::class) {
```
aaron-steinfeld, line 74: upload archives is the deprecated publishing mechanism - any reason we can't use the new one? For that matter, most of this looks like it's just changing the destination used by our regular publish plugin - can we use (or modify that)?  This isn't the kind of stuff we want copied around repo to repo.

aaron-steinfeld, line 74: Are we intentionally uploading every subproject? Any way for them to opt out? Using subprojects is a big gradle antipattern because it makes it very hard to understand how projects are configuring and easy to mess up - that goes doubly for publishing - we don't want to add a new project and have it automatically start publishing to maven central.

ravisingal: I've followed the official documentation for this task: https://central.sonatype.org/pages/gradle.html

ravisingal: we can try moving this out to individual sub-projects. I added here to avoid duplicates.

aaron-steinfeld: That's horribly outdated. The upload mechanism was deprecated way back in gradle 1 in 2014 - its remained for legacy support (probably this documentation 🙄 . There's not too much to  using the new mechanism, I can help you out.

See example here: https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:complete_example

ravisingal:
closing this in favor of https://github.com/hypertrace/javaagent/pull/285

PR #285, opened 2021-03-11
Publish java agent to maven central

ravisingal:


.circleci/config.yml
```diff
@@ -107,7 +107,9 @@ workflows:
             tags:
               only: /.*/
       - release-bintray-publish:
-          context: hypertrace-publishing
+          context:
+            - hypertrace-publishing
```
aaron-steinfeld, line 111: do we still need the old context? Also, are there plans on moving this repo off circleci? i think it's the last one preventing us from completely disconnecting cci.

ravisingal: This context has `DOCKER_USERNAME` and `DOCKER_PASSWORD`. Not sure if we need `docker login` here.

will move this repo to Github Actions as part of another PR (maybe next week).

.circleci/config.yml
```diff
@@ -99,15 +99,19 @@ workflows:
   build-and-publish:
     jobs:
       - build:
+          context: maven-central-publish
```
aaron-steinfeld, line 102: we should only add this on the release job. If the plugin requires this when publish isn't invoked, we should update that on the plugin side (as users will not have these values set)

.circleci/config.yml
```diff
@@ -99,15 +99,19 @@ workflows:
...
           filters:
             tags:
               only: /.*/
       - muzzle:
+          context: maven-central-publish
           filters:
             tags:
               only: /.*/
       - release-bintray-publish:
```
aaron-steinfeld, line 111: this job is no longer going to bintray so should be renamed

.circleci/config.yml
aaron-steinfeld: I see it's setting up and logging into docker, but I can't imagine why. Guessing that was just copy pasted from a job building docker images - this one isn't, right?

ravisingal: This repo also builds docker image: https://github.com/hypertrace/javaagent/blob/main/Makefile#L22-L31

It is used for copying java agent to container filesystem before starting the java process using a mutating webhook.

.circleci/config.yml
ravisingal: done.

.circleci/config.yml
aaron-steinfeld: I dare not ask why we're using make here..

.circleci/config.yml
ravisingal: Raised the following PR: https://github.com/hypertrace/hypertrace-gradle-publish-plugin/pull/16

aaron-steinfeld (approved):

PR #287, opened 2021-03-15
fix: use the correct otlp endpoint

tim-mwangi:
## Description
Based on https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#otlp-exporter-both-span-and-metric-exporters.
Neither OTEL_EXPORTER_OTLP_TRACES_ENDPOINT or otel.exporter.otlp.traces.endpoint work

### Testing
Ran it on a k8s sample application to test for the case where the default `http://localhost:4317` is not set automatically and hence hides the bug.

### Checklist:
- [✅  ] My changes generate no new warnings
- [✅  ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/HypertraceAgentConfiguration.java
+++ b/otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/HypertraceAgentConfiguration.java
@@ -41,6 +41,7 @@ public class HypertraceAgentConfiguration implements PropertySource {
   private static final String OTEL_PROCESSOR_BATCH_MAX_QUEUE = "otel.bsp.max.queue.size";
   private static final String OTEL_DEFAULT_LOG_LEVEL =
       "io.opentelemetry.javaagent.slf4j.simpleLogger.defaultLogLevel";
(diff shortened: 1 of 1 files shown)


pavolloffay (approved):

tim-mwangi:
@pavolloffay @shashank11p the smoke tests are failing since `otel.library.version` does not match the agent version. Also I debugged a bit and this attribute is not even set. Should we be concerned about this?

eg. SpringBootDisabledBodyCaptureSmokeTest line 83
```
Assertions.assertTrue(
        getSpanStream(traces)
                .flatMap(span -> span.getAttributesList().stream())
                .filter(attribute -> attribute.getKey().equals(OTEL_LIBRARY_VERSION_ATTRIBUTE))
                .map(attribute -> attribute.getValue().getStringValue())
                .filter(value -> value.equals(currentAgentVersion))
                .count()
            > 0)
```

tim-mwangi:
I dug deeper into this. It seems as if there was a fix put in recently, https://github.com/open-telemetry/opentelemetry-java/commit/9d96758195949bb326bc819805e83c519297ca65, to also check the `otel.exporter.otlp.traces.endpoint` config and fall back to the value of `otel.exporter.otlp.endpoint`. The version we are currently using only supports the latter.

PR #293, opened 2021-03-22
Fix config being loaded twice

### LINKED ISSUE #223: Hypertrace config is being loaded twice
The config file is being loaded twice. The config should live in the bootstrap classloader and be loaded only once.

```
~/projects/demo-apps/vertx-examples/web-examples(4.x) » JAVA_TOOL_OPTIONS=-javaagent:/Users/ploffay/Downloads/hypertrace-agent-all.jar HT_SERVICE_NAME=vertx-rest-demo vertx run src/main/java/io/vertx/example/web/rest/SimpleREST.java                        130 ↵ ploffay@Pavols-MBP
Picked up JAVA_TOOL_OPTIONS: -javaagent:/Users/ploffay/Downloads/hypertrace-agent-all.jar
[main] INFO org.hypertrace.agent.core.config.HypertraceConfig - Config loaded: {"serviceName":"vertx-rest-demo","reporting":{"endpoint":"http://localhost:9411/api/v2/spans","opa":{"endpoint":"http://opa.traceableai:8181/","pollPeriodSeconds":30}},"dataCapture":{"httpHeaders":{"request":true,"response":true},"httpBody":{"request":true,"response":true},"rpcMetadata":{"request":true,"response":true},"rpcBody":{"request":true,"response":true},"bodyMaxSizeBytes":131072},"propagationFormats":["TRACECONTEXT"]}
[opentelemetry.auto.trace 2021-01-11 11:07:42:881 +0100] [main] INFO org.hypertrace.agent.core.config.HypertraceConfig - Config loaded: {"serviceName":"vertx-rest-demo","reporting":{"endpoint":"http://localhost:9411/api/v2/spans","opa":{"endpoint":"http://opa.traceableai:8181/","pollPeriodSeconds":30}},"dataCapture":{"httpHeaders":{"request":true,"response":true},"httpBody":{"request":true,"response":true},"rpcMetadata":{"request":true,"response":true},"rpcBody":{"request":true,"response":true},"bodyMaxSizeBytes":131072},"propagationFormats":["TRACECONTEXT"]}
```

ryandens:
## Description
Resolves #223 (kind of). This bug no longer seems to be present, but as a result of the investigation into #223 a small thread-safety issue was noticed in the double-checked locking implementation. To fix it, I pulled the implementation of memoizing the `AgentConfig` into a separate class. This made it easier to inject a test double (specifically, a spy) to verify that the config is properly memoized. 

### Testing
Added test cases on the memoization and resetting of the memoized value. 

### Checklist:

### Documentation
No new documentation required

This solution was based off of/inspired by 
- [The "Double-Checked Locking is Broken Declaration"](https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html) 
- [JSR 133 FAQ "What does volatile do?"](http://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html#volatile)
- [Guava's NonSerializableMemoizingSupplier](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Suppliers.java#L150) reference implementation


javaagent-core/src/test/java/org/hypertrace/agent/core/config/AgentConfigSupplierTest.java
```diff
@@ -0,0 +1,100 @@
...
+    assertNotSame(agentConfig, secondConfig);
+  }
+
+  /**
+   * thread-safe {@link Config.AgentConfig} {@link Supplier} that also counts how many times it has
+   * been invoked
+   */
+  private static final class CountingAgentConfigSupplier implements Supplier<Config.AgentConfig> {
+
+    private final AtomicInteger count = new AtomicInteger(0);
```
ryandens, line 83: This `AtomicInteger` is probably overkill, but I left it in on the off chance someone wants to expand on this test case later. I originally wrote this test to try to reproduce the double-checked locking race condition that can occur by using a multi-threaded scheduled executor service to ask for the `AgentConfig` 10,000 times via separate callables all calling `get` on the same `AgentConfigSupplier` instance and verifying that the inner `CountingAgentConfigSupplier` was only invoked once. I wasn't able to reproduce the bug in that fashion, so I pulled out the test case and simply tested the supplier "the normal way". Since `CountingAgentConfigSupplier` is initialized `BeforeEach`, there isn't necessarily a need to have this class be thread-safe and we could simply make this an `int`.  Happy to back it out to avoid complexity or keep it in if we want to expand the test cases around this more.

javaagent-core/src/main/java/org/hypertrace/agent/core/config/HypertraceConfig.java
```diff
@@ -51,7 +51,8 @@ private HypertraceConfig() {}
   // so avoiding for perf reasons
   private static volatile boolean servletCausingException;
 
-  private static AgentConfig agentConfig;
```
pavolloffay, line 54: Thanks for putting this together. I like the testability, but on the other hand, I would like to keep it as simple as possible. Would it be enough to just use `volatile` on this property and not use the supplier approach?

ryandens: Yeah, no problem! I think in addition to making this field `volatile`, we would also want to synchronize the write to this field in `HypertraceConfig.reset()` unless you're not worried about this class being thread-safe when APIs exposed for testing purpose are in use.

pavolloffay (approved):

pavolloffay:
thanks @ryandens

PR #299, opened 2021-04-06
Add endpoint sampler

pavolloffay:
TODOs:

- [x] Add util to parse HT tracestate
- [x] Define HT tracestate  https://github.com/hypertrace/specification/pull/7
- [x] update servlet instrumentation to do not capture additional data in core mode

This PR updates only servlet instrumentation to respect capture mode in the tracestate

Depends on OTEL 1.1.0 release that adds sampler SPI and passes URLs to sampler.


pavolloffay:
The smoke test fails bc the endoint sampler is not initialized in the final java agent distribution.

PR #305, opened 2021-04-13
Default to zipkin exporter

shashank11p:
Revert defaulting to OTLP exporter.


smoke-tests/src/test/java/org/hypertrace/agent/smoketest/SpringBootSmokeTest.java
```diff
@@ -79,45 +79,33 @@ public void get() throws IOException, InterruptedException {
...
-    Assertions.assertEquals(
-        ResourceAttributes.TELEMETRY_AUTO_VERSION.getKey(),
-        traces.get(0).getResourceSpans(0).getResource().getAttributes(10).getKey());
-    Assertions.assertEquals(
-        ResourceAttributes.TELEMETRY_SDK_LANGUAGE.getKey(),
-        traces.get(0).getResourceSpans(0).getResource().getAttributes(11).getKey());
-    Assertions.assertEquals(
-        ResourceAttributes.TELEMETRY_SDK_NAME.getKey(),
-        traces.get(0).getResourceSpans(0).getResource().getAttributes(12).getKey());
+        traces.get(0).getResourceSpans(0).getResource().getAttributes(0).getKey());
```
davexroth, line 84: Rather than revert these tests, can we configure the tests to run with otlp?

shashank11p: Sure, we can use OTLP in config while running these tests.

davexroth (approved):

PR #311, opened 2021-04-22
Instrument vertx client directly

### LINKED ISSUE #310: Vertx client instrumentation is broken in 1.1
Fix will require instrumenting vertx client directly (instead of Netty as we do now).

Disabled test: https://github.com/hypertrace/javaagent/pull/308/files#diff-7bf0cebf0d08303f88f759ed4f34c318c1dc246cf8bb787a313a45cdb73b6576R33

pavolloffay:
Resolves #310

### DIFF
--- a/instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/HttpRequestInstrumentation.java
+++ b/instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/HttpRequestInstrumentation.java
@@ -1,0 +1,271 @@
+/*
+ * Copyright The Hypertrace Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.javaagent.instrumentation.hypertrace.vertx;
+
+import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.AgentElementMatchers.implementsInterface;
+import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.ClassLoaderMatcher.hasClassesNamed;
+import static net.bytebuddy.matcher.ElementMatchers.is;
+import static net.bytebuddy.matcher.ElementMatchers.isMethod;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.javaagent.instrumentation.api.CallDepthThreadLocalMap;
+import io.opentelemetry.javaagent.instrumentation.api.ContextStore;
+import io.opentelemetry.javaagent.instrumentation.api.InstrumentationContext;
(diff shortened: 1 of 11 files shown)


instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/HttpRequestInstrumentation.java
```diff
@@ -0,0 +1,261 @@
...
+          InstrumentationContext.get(HttpClientRequest.class, Contexts.class).get(request);
+      if (contexts == null) {
+        return;
+      }
+      Span span = Span.fromContext(contexts.context);
+
+      BoundedCharArrayWriter buffer =
+          InstrumentationContext.get(MultiMap.class, BoundedCharArrayWriter.class)
+              .get(request.headers());
+      if (buffer != null) {
```
shashank11p, line 106: Dont we want to check if 
```
(HypertraceConfig.get().getDataCapture().getHttpBody().getRequest().getValue()
          && ContentTypeUtils.shouldCapture(contentType))

File: instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/HttpRequestInstrumentation.java [hunk 2]
```diff
@@ -0,0 +1,261 @@
...
+          && ContentTypeUtils.shouldCapture(contentType)) {
+
+        ContextStore<MultiMap, BoundedCharArrayWriter> contextStore =
+            InstrumentationContext.get(MultiMap.class, BoundedCharArrayWriter.class);
+        BoundedCharArrayWriter buffer = contextStore.get(request.headers());
+        if (buffer == null) {
+          buffer = BoundedBuffersFactory.createWriter();
+          contextStore.put(request.headers(), buffer);
+        }
+        buffer.write(chunk);
```
shashank11p, line 220: Why are we calling this here and in EndRequestAdvice both?

instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/ResponseBodyWrappingHandler.java
```diff
@@ -0,0 +1,56 @@
...
+  public ResponseBodyWrappingHandler(Handler<Buffer> wrapped, Span span) {
+    this.wrapped = wrapped;
+    this.span = span;
+  }
+
+  @Override
+  public void handle(Buffer event) {
+    String responseBody = event.getString(0, event.length());
+    if (span.isRecording()) {
+      span.setAttribute(HypertraceSemanticAttributes.HTTP_RESPONSE_BODY, responseBody);
```
shashank11p, line 44: Should we also check if we should capture response body?

instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/HttpRequestInstrumentation.java
pavolloffay: yes, this writes a chunk and in the end method it as well writes a chunk and sends the request.

instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/HttpRequestInstrumentation.java
pavolloffay: we don't have to check it here bc the buffer is null if the body shouldn't be captured.

instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/ResponseBodyWrappingHandler.java
pavolloffay: not needed as this is reached only after the check from the previous instrumentation point.

PR #312, opened 2021-04-23
Snyk

ryandens:
## Description
Upgrades dependencies to use newer, non-vulnerable versions or more tightly scoped Gradle configurations where applicable. In order to minimize the scope of this change, dependencies were only updated for projects that had dependencies flagged by the `snyk test` command

The most significant refactor is of the `testing-common` project. Specifically, Jetty 8 is old enough that there are vulnerabilities that will only be fixed in newer major versions. This isn't particularly a problem though, as this `java-library` is not published externally. We restrict the `snyk test` command to only look for dependencies on the `runtimeClasspath`. In order to make it more clear the intention of this project, I refactored it to use the `java-test-fixtures` plugin and changed the scope of the dependencies used in the project to only be on the `testFixturesRuntimeClasspath` resulting in a better separation of testing dependencies from runtime dependencies. This allows us to safely keep using Jetty 8 without worrying about accidentally including it in a release in the future. 

### Testing
The risk of the `testing-common` changes affecting a production release is very small, since it is only consumed by the `testImplementation` configuration in other projects. The AutoService upgrades are best checked by our smoke tests across various runtimes/application servers, but i also manually verified that contents of `META-INF/services/io.opentelemetry.javaagent.tooling.InstrumentationModule` was the same. 

I think riskiest change was the upgrade of  `protobuf-java-util` from `3.11.4` to `3.15.8`. Based on our smoke tests that interact with the otel collector/backend which receive messages serialized/deserialized with this util, I feel pretty confident. But, I'm not an expert with protobuf so if there's any "gotchas" that I should look for I'm happy to chase them down

Here's the snyk vulnerability report
```bash

Testing /Users/rdens/git/hypertrace/javaagent...

Tested 24 dependencies for known issues, found 1 issue, 1 vulnerable path.


Issues to fix by upgrading:

  Upgrade com.google.protobuf:protobuf-java-util@3.11.4 to com.google.protobuf:protobuf-java-util@3.15.0 to fix
  ✗ Information Disclosure [Medium Severity][https://snyk.io/vuln/SNYK-JAVA-COMGOOGLEGUAVA-1015415] in com.google.guava:guava@28.1-android
    introduced by com.google.protobuf:protobuf-java-util@3.11.4 > com.google.guava:guava@28.1-android



Organization:      hypertrace
Package manager:   gradle
Target file:       build.gradle.kts
Project name:      javaagent/javaagent-core
Open source:       no
Project path:      /Users/rdens/git/hypertrace/javaagent
Licenses:          enabled


```

### Checklist:

### Documentation
N/A

### Links
- https://snyk.io/vuln/SNYK-JAVA-COMGOOGLEGUAVA-1015415
We weren't using guava in any significant way on the runtime classpath, so there's no significant exploit risk here. But, we definitely should still update it.


instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/HttpRequestInstrumentation.java
```diff
@@ -68,7 +68,10 @@
         HttpRequestInstrumentation.class.getName() + "$WriteRequestAdvice_string");
     transformers.put(
         isMethod()
-            .and(named("write").and(takesArguments(1)).and(takesArgument(0, is(Buffer.class)))),
+            .and(
+                named("write")
+                    .and(takesArguments(1))
+                    .and(takesArgument(0, named("io.vertx.core.buffer.Buffer")))),
```
pavolloffay, line 74: why this change? I would prefer to keep it as a type instead of a string.

instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/HttpRequestInstrumentation.java
```diff
@@ -78,7 +81,11 @@
         isMethod().and(named("end").and(takesArgument(0, is(String.class)))),
         HttpRequestInstrumentation.class.getName() + "$EndRequestAdvice_string");
     transformers.put(
-        isMethod().and(named("end").and(takesArguments(1)).and(takesArgument(0, is(Buffer.class)))),
+        isMethod()
+            .and(
+                named("end")
+                    .and(takesArguments(1))
+                    .and(takesArgument(0, named("io.vertx.core.buffer.Buffer")))),
```
pavolloffay, line 88: the same as above.

buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
```diff
@@ -24,6 +26,18 @@
   @Override
   public void apply(Project project) {
     project.getPlugins().apply(JavaLibraryPlugin.class);
+    final Configuration libraryConfiguration = project.getConfigurations()
```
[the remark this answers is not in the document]
ryandens: Good call, definitely agreed a comment is needed here!

instrumentation/vertx-web-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/vertx/HttpRequestInstrumentation.java
ryandens: When I removed the vertx dependency from the `runtimeClasspath`, the `byteBuddyJava` task starts failing with a `java.lang.NoClassDefFoundError: io/vertx/core/buffer/Buffer`. This surprised me because I assumed that this task leveraged the `compileClasspath` rather than the `runtimeClasspath`, but perhaps it builds its own classpath derived from the `runtimeClasspath`? When I went looking at other projects in this repo, I noticed that this project was the only one to use the class literals in the `transformers`, which is why the other projects didn't have a problem with removing the library being instrumented from the `runtimeClasspath`.


I too would prefer the type instead of the `String`. Writing up this comment made me realize that the reason removing vertx from the `runtimeClasspath` causes the `byteBuddyJava` task to fa
il is because we scope the `ByteBuddyPluginConfigurator` to only use the `runtimeClasspath` as shown here:

```kotlin
afterEvaluate{
    io.opentelemetry.instrumentation.gradle.bytebuddy.ByteBuddyPluginConfigurator(project,
            sourceSets.main.get(),
            "io.opentelemetry.javaagent.tooling.muzzle.collector.MuzzleCodeGenerationPlugin",
            project(":javaagent-tooling").configurations["instrumentationMuzzle"] + configurations.runtimeClasspath
    ).configure()
}
```

If instead we use `configurations.compileClasspath` rather than `configurations.runtimeClasspath` to configure the classpath for the bytebuddy gradle plugin. But, I'm not sure if that will have any other negative impacts. I looked closer at the open telemetry java nstrumentation project and saw that they also restricted the byteBuddyJava classpath to the tooling and runtime classpaths. As a result, the parts of their code that execute during the `byteBuddyJava` class use string literals for class names rather than the actual types, such as with their [Vertx Web HttpRequestInstrumentation class](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/0b0516bd1a0599c8f536cc6b2151782000681f39/instrumentation/vertx-web-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vertx/client/HttpRequestInstrumentation.java#L48)

ryandens: Following up, while I too would prefer to use the type rather than the `String` name, I also am weary of configuring bytebuddy to use the compileClasspath. As a result, I think we should follow the standard set by otel and keep these as Strings

pavolloffay: sounds good, thanks for the explanation :)

pavolloffay (approved):

PR #319, opened 2021-05-12
Ignore: fix classes in the bootstrap classloader

pavolloffay:
Ignore: just for archive


pavolloffay:
already done in another PR

PR #321, opened 2021-06-10
✨  Capture HTTP bodies on Undertow

ryandens:
### Background
- Undertow is a web server written in Java
- It's leveraged by/integrated with several large projects including Wildfly and RESTEasy. 
- Projects that build on top of Undertow often leverage its Java EE Servlet API Implementation
- Its implementation of the Servlet API does not always leverage the Servlet API `javax.servlet.http.HttpServletRequest.getInputStream()` 

### Overview
- ✨ Adds Instrumentation to associate `org.xnio.channels.StreamSourceChannel` instances with a `SpanAndBuffer` instance to give the proper context
- ✨ Adds instrumentation to copy data from a `ByteBuffer` after it is read from an `org.xnio.channels.StreamSourceChannel` into a `ByteBuffer` and associate that data with the request body Span attribute. 

### Testing
- ✅ Adds a test case that reproduces a situation where our existing implementation previously did not capture request body on an Undertow deployment that we would otherwise expect to work on
- Manual tests on a Wildfly servlet container with a RESTEasy JAX-RS deployment 

### Checklist:

### Documentation
No new documentation is required, as we broadly claim Java Servlet support of which this undertow use-case is an implementation. This just happens to be a corner case.


instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
```diff
@@ -0,0 +1,89 @@
...
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void enter(
+        @Advice.Enter int callDepth,
+        @Advice.Return final int numBytesRead,
+        @Advice.This final StreamSourceChannel streamSourceChannel,
+        @Advice.Argument(0) final ByteBuffer byteBuffer) {
+      if (callDepth > 0) {
+        return;
+      }
+      HypertraceCallDepthThreadLocalMap.reset(StreamSourceChannel.class);
```
pavolloffay, line 80: if an exception is thrown in the method the thread map will not be reset.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
```diff
@@ -0,0 +1,89 @@
...
+      final ContextStore<StreamSourceChannel, SpanAndObjectPair> contextStore =
+          InstrumentationContext.get(StreamSourceChannel.class, SpanAndObjectPair.class);
+      final SpanAndObjectPair spanAndObjectPair = contextStore.get(streamSourceChannel);
```
pavolloffay, line 83: nit: remove `final` it does not help here and it just adds noise.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/UndertowHttpServerExchangeInstrumentation.java
```diff
@@ -0,0 +1,80 @@
...
+        named("getRequestChannel")
+            .and(takesArguments(0))
+            .and(returns(named("org.xnio.channels.StreamSourceChannel")))
+            .and(isPublic()),
+        UndertowHttpServerExchangeInstrumentation.class.getName() + "$getRequestChannel_advice");
+    return Collections.unmodifiableMap(matchers);
+  }
+
+  /** TODO */
+  static final class getRequestChannel_advice {
```
pavolloffay, line 64: nit: `GetRequest...`

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/UndertowHttpServerExchangeInstrumentation.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+    public static void exit(
+        @Advice.This final HttpServerExchange thizz,
+        @Advice.Return final StreamSourceChannel returnedChannel) {
+      final Context serverContext = UndertowHttpServerTracer.tracer().getServerContext(thizz);
```
pavolloffay, line 70: again `final` does not help here. It's just noise for the reader

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/UndertowHttpServerExchangeInstrumentation.java
```diff
@@ -0,0 +1,80 @@
...
+      final Span span = Span.fromContext(serverContext);
+      final SpanAndObjectPair spanAndObjectPair =
+          new SpanAndObjectPair(Objects.requireNonNull(span, "span should not be null"));
```
pavolloffay, line 73: `Objects.requireNonNull(span, "span should not be null")` span is never null. `Span.fromContext(serverContext)` returns either recording span or default one.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
```diff
@@ -0,0 +1,63 @@
...
+    try {
+      body = boundedByteArrayOutputStream.toStringWithSuppliedCharset().trim();
+    } catch (UnsupportedEncodingException e) {
+      throw new IllegalArgumentException("illegal encoding", e);
+    }
+    log.info(body);
+    spanAndObjectPair.getSpan().setAttribute(HypertraceSemanticAttributes.HTTP_REQUEST_BODY, body);
+  }
+
+  private static final Logger log = LoggerFactory.getLogger(Utils.class);
```
pavolloffay, line 62: the logger is in this project defined at the top of the class.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
```diff
@@ -0,0 +1,63 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class Utils {
+
+  public static void handleTransfer(
+      final ByteBuffer byteBuffer, int numBytesRead, SpanAndObjectPair spanAndObjectPair) {
+    final Object associatedObject = spanAndObjectPair.getAssociatedObject();
+    final HttpServerExchange exchange;
+    if (associatedObject instanceof HttpServerExchange) {
```
pavolloffay, line 36:
```java
if (!(associatedObject instanceof HttpServerExchange)) {
  return;
}

exchange = (HttpServerExchange) associatedObject;
```

is more compact with less branching factor

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
```diff
@@ -0,0 +1,63 @@
...
+    if (!exchange.isRequestComplete()) {
+      return;
+    }
+    final BoundedByteArrayOutputStream boundedByteArrayOutputStream =
+        BoundedBuffersFactory.createStream(
+            byteBuffer.remaining(), Charset.forName(exchange.getRequestCharset()));
+    final byte[] dst = new byte[byteBuffer.remaining()];
+    byteBuffer.get(dst);
+    boundedByteArrayOutputStream.write(dst, 0, dst.length);
+    log.info("test");
```
pavolloffay, line 51: remove?

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
```diff
@@ -0,0 +1,63 @@
...
+    final String body;
+    try {
+      body = boundedByteArrayOutputStream.toStringWithSuppliedCharset().trim();
```
pavolloffay, line 54: why trim? We do not call it anywhere else

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
```diff
@@ -0,0 +1,63 @@
...
+    } catch (UnsupportedEncodingException e) {
+      throw new IllegalArgumentException("illegal encoding", e);
```
pavolloffay, line 56: can this happen? wasn't the correct encoding parsed before?

pavolloffay (commented):
good progress 💯 

A couple of things to look at:
* avoid using `final` for local variables in short methods. 
* the body capture should for content type and capture only  specific types - see the spec in the HT org. Also check for content length to preallocate size of the buffer and charset. 
* the instrumentation code should use HT config to check if capture is enabled.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
ryandens: Yeah, this PR is definitely still a "draft" and this, in my opinion, is mostly what's holding it back. The `ByteBuffer` that is written to by the `org.xnio.channels.StreamSourceChannel.read(java.nio.ByteBuffer)` in my test is read twice. The first time, it reads in the 8 chars that make up the request body and puts them into the `java.nio.ByteBuffer`. The second time, no bytes are read (read returns -1). Right now, how the implementation works, is that we ignore reads until `exchange.isRequestComplete()` is true. Then, we read the bytes into from the ByteBuffer` into a `ByteArray`. However, we don't know how many bytes should actually be read in, since `numBytesRead` is `-1`. So, it reads the entire byte buffer, including null bytes, into the byte array and then writes that to the bounded stream, resulting in a reported request body like `echo=bar....` where `` is repeated for the entire remaining size of the `ByteBuffer`.

This is definitely not ideal. I'm just wrapping up a solution to instead capture the bytes only when they are written to the `ByteBuffer`. In this case, we create a read-only ByteBuffer and reset the position of the read-only ByteBuffer to where it was before the read, like this: 
```java
readOnlyBuffer.position(readOnlyBuffer.position() - numBytesRead); 
final byte[] dst = new byte[numBytesRead];
readOnlyBuffer.get(dst);
boundedByteArrayOutputStream.write(dst, 0, dst.length);
```

Then, we can remove this `trim` and save on the space we allocate in the byte array. In my example test, `numBytesRead` is `8` whereas buffer.getRemaining()` is ~16,000.

What are your thoughts on this solution? is there a better way to capture the contents of the ByteBuffer without modifying the ByteBuffer instance used by undertow? 

I'm wondering if we'll need to modify my existing implementation to put a `ByteBufferSpanPair` in the instrumentation context (like the servlet implementation) when `getRequestChannel` is invoked, using the `StreamSourceChannel.read` as the key Then, we can look up the that  `ByteBufferSpanPair` when StreamSourceChannel.read is invoked. Finally, we can add a `ExchangeCompletionListener` to the `HttpExchange` to convert the `BoundedByteArrayOutputStream` object to a `String` and set the `HTTP_REQUEST_BODY` attribute on the associated span. This would harden our implementation of request body capturing in the case where `StreamSourceChannel.read` is invoked multiple times. I'm not quite sure yet if this is a reasonable fear 🤔

pavolloffay: > This is definitely not ideal. I'm just wrapping up a solution to instead capture the bytes only when they are written to the ByteBuffer. In this case, we create a read-only ByteBuffer and reset the position of the read-only ByteBuffer to where it was before the read, like this:

All other instrumentations hook directly into buffer write methods - e.g. see java stream instrumentation. Definitely avoid wrapping of buffers :)

ryandens: Ah okay! Just to make sure I understand, you think we should instrument `java.nio.ByteBuffer.put` APIs and to write the provided `bytes`/`byte[]` to the `BoundedByteArrayOutputStream` if we find that the `ByteBuffer` instance was associated with a request body? Meaning we would: 

1. put the the `SpanAndBuffer` in the `InstrumentationContext` with the `StreamSourceChannel` instance as the key when `HttpServerExchange.getRequestChannel()` is invoked
1. put the the `SpanAndBuffer` in the `InstrumentationContext` with the `ByteBuffer` instance as the key when `StreamSourceChannel.read(ByteBuffer)` is invoked on a `StreamSourceChannel` in the `InstrumentationContext`
1.  Write to the `BoundedByteArrayOutputStream` when the `ByteBuffer.put` APIs are called on a `ByteBuffer` instance in the `InstrumentationContext`

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
ryandens: Yeah, you're correct. This exception is propagated by `toStringWithSuppliedCharset`. I think I should modify `toStringWithSuppliedCharset` to catch the `UnsupportedEncodingException` and wrap it in an unchecked exception, since `toStringWithSuppliedCharset` calls `toString(String)` with the charset name retrieved from a `Charset` instance.  So
```
  public synchronized String toStringWithSuppliedCharset() throws UnsupportedEncodingException {
    return super.toString(this.charset.name());
  }
```
  
 Would  become: 
```
public synchronized String toStringWithSuppliedCharset() {
    try {
      return super.toString(this.charset.name());
    } catch (UnsupportedEncodingException e) {
      throw new IllegalStateException("Charset name is guaranteed to be valid");
    }
  }
```

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
```diff
@@ -0,0 +1,89 @@
...
+import net.bytebuddy.matcher.ElementMatcher.Junction;
+import org.hypertrace.agent.core.instrumentation.HypertraceCallDepthThreadLocalMap;
+import org.hypertrace.agent.core.instrumentation.SpanAndObjectPair;
+import org.xnio.channels.StreamSourceChannel;
+
+public final class StreamSourceChannelInstrumentation implements TypeInstrumentation {
+
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    return AgentElementMatchers.safeHasSuperType(named("org.xnio.channels.StreamSourceChannel"));
```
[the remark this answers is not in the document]
ryandens: Nope! OTEL doesn't instrument any `xnio` types or the implementations of it in Undertow as far as I can tell.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
pavolloffay: yes something like this, again the idea is to minimize the intrusiveness of our code. Also note that some frameworks use buffers that can be read multiple times or the buffer provides interceptors API, in this case it's preferrable to use these APIs.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
pavolloffay: I would rather avoid throwing the exception and instead just add a log. The agent should not break the app in any case. 

On a different note I don't recall why the method is synchronized.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
ryandens: Just ran into something interesting with this approach! In the case of this use case on Undertow, we see this sequence of events the first time the request body is read:

### Approach 1: instrument `ByteBuffer` APIs, watching for `.put(....)` invocations on the `ByteBuffer` passed to `StreamSourceChannel.read(ByteBuffer)`.=
1. `HttpServerExchange.getRequestChannel()` is invoked. Our instrumentation properly puts the `StreamSourceChannel` in the `ContextStore<StreamSourceChannel, SpanAndObjectPair>` with the span
2. `StreamSourceChannel.read(ByteBuffer)` is called on the `StreamSourceChannel` instance returned from `HttpServerExchange.getRequestChannel()`. Now, we can put the `ByteBuffer` in a `ContextStore<ByteBuffer, SpanAndObjectPair>`
3. In the scope of `StreamSourceChannel.read(ByteBuffer)`, undertow delegates to a chain of object in order to properly read the request body from the `StreamSourceChannel` into the provided `ByteBuffer`. In the use-case of my test, this chain looks like [HttpServerExchange$ReadDispatchChannel.read(ByteBuffer)](https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/server/HttpServerExchange.java#L2419) -> [DetachableStreamSourceChannel.read(ByteBuffer)](https://github.com/undertow-io/undertow/blob/77c3b9bec4895f48aee3155747cefb9dccd2ec6b/core/src/main/java/io/undertow/channels/DetachableStreamSourceChannel.java#L209) -> [ConduitStreamSourceChannel.read(ByteBuffer)](https://github.com/xnio/xnio/blob/3.x/api/src/main/java/org/xnio/conduits/ConduitStreamSourceChannel.java#L127) -> [FixedLengthStreamSourceConduit.read(ByteBuffer)](https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/conduits/FixedLengthStreamSourceConduit.java#L244) -> [ReadDataStreamSourceConduit.read(ByteBuffer)](https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/conduits/ReadDataStreamSourceConduit.java#L60).
4. `ReadDataStreamSourceConduit.read(ByteBuffer)` gets the bytes associated with the request body in the form of a `PooledByteBuffer` from the `AbstractServerConnection` API `getExtraBytes()`. This `PooledByteBuffer` composes the src `ByteBuffer` and uses `ByteBuffer.put(ByteBuffer)` API to write to the buffer we've instrumented
5. `java.nio.ByteBuffer.put(ByteBuffer)` delegates to `java.nio.ByteBuffer.put(byte)` for each `byte` in the src `ByteBuffer`. Unfortunately, not all `ByteBuffer`s use this implementation. Notably, the `ByteBuffer` instance here is a `java.nio.DirectByteBuffer`, which overrides the `put(ByteBuffer)` API and used `Unsafe` to directly copy the memory from the source `ByteBuffer` to the destination `ByteBuffer`, so instrumentation of the other `ByteBuffer.put(...)` APIs will not help us capture the request body 


### Approach 2: Instrument `ByteBuffer`, watching for `.put(....)` invocations on the `ByteBuffer` used by the Undertow `HttpReadListener`

1. This led me down the other path, capture the request body when it is being read into a `ByteBuffer` the very first time by undertow in [HttpReadListener](https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/server/protocol/http/HttpReadListener.java#L158). 
2. The above line delegates to the [ConduitStreamSourceChannel.read(ByteBuffer)](https://github.com/xnio/xnio/blob/3.x/api/src/main/java/org/xnio/conduits/ConduitStreamSourceChannel.java#L127) in xnio followed by [NioSocketConduit.read(ByteBuffer)](https://github.com/xnio/xnio/blob/3.x/nio-impl/src/main/java/org/xnio/nio/NioSocketConduit.java#L286). 
3. `NioSocketConduit` then delegates to the `java.nio.channels.SocketChannel` implementation [sun.nio.ch.SocketChannelImpl.read(ByteBuffer)](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/sun/nio/ch/SocketChannelImpl.java#L417). 
4. The above line uses a `NativeDispatcher` to natively read from the `SocketChannelImpl` into the destination `ByteBuffer`, again sidestepping the public `ByteBuffer` APIs.

Given those two situations, I don't think that watching for `.put(..)` invocations on a `ByteBuffer` will help us the same way it did for `InputStream`s. I think no matter what, we're going to have to look at the `ByteBuffer` instance after `read` has occurred and pull the request body information out of it based on how many bytes were read. Right now, the instrumentation we have applied in this PR captures the body in the context of the `HttpServerExchange.ReadDispatchChannel.read()` API, but I think we would do better to have it captured in the context of the read in `HttpReadListener.handleEvent` because undertow goes through significant lengths to make sure that it is only invoked once per request. Functionally though, the approach would be pretty similar to what we have now

ryandens: > All other instrumentations hook directly into buffer write methods - e.g. see java stream instrumentation. Definitely avoid wrapping of buffers :)

Also, circling back on this, note that the approach implemented here doesn't wrap the buffer in any way. We instrument the `org.xnio.channels.StreamSourceChannel.read(ByteBuffer)` API to capture the request body on method exit after it has been written to the `ByteBuffer`. In order to prevent modification of the destination `ByteBuffer`, before reading the data out from the `ByteBuffer` into our `BoundedByteArrayOutputStream`, I call `ByteBuffer.asReadOnlyBuffer` as a defensive measure to prevent modification of the original. 

```java

    @Advice.OnMethodExit(suppress = Throwable.class)
    public static void enter(
        @Advice.Enter int callDepth,
        @Advice.Return final int numBytesRead,
        @Advice.This final StreamSourceChannel streamSourceChannel,
        @Advice.Argument(0) final ByteBuffer byteBuffer) {
      if (callDepth > 0) {
        return;
      }
      HypertraceCallDepthThreadLocalMap.reset(StreamSourceChannel.class);
      final ContextStore<StreamSourceChannel, SpanAndBuffer> contextStore =
          InstrumentationContext.get(StreamSourceChannel.class, SpanAndBuffer.class);
      final SpanAndBuffer spanAndBuffer = contextStore.get(streamSourceChannel);
      if (spanAndBuffer != null) {
        Utils.handleRead(byteBuffer.asReadOnlyBuffer(), numBytesRead, spanAndBuffer);
      }
    }
```

Note here that the original `ByteBuffer` is not modified, just that the instance we pass to `Utils.handleRead` for analysis is designated read-only to prevent modification

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
ryandens: Works for me! 👍 Yeah, I'm not sure why it is either, I'll look into it in the future but for now I'm going to hold off on changing it since some of our users are waiting on the instrumentation fix for undertow 😃

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
ryandens: I disagree here - I find that defaulting local variables to final discourages mutating references to objects and results in code that is easier to read and maintain.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
ryandens: Thanks! Good catch 🎉

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
pavolloffay: `final` is useful in large methods where there is a higher probability of making a mistake by accidentally reusing the variable. The `final` should be used to improve the readability of large code blocks or as a "documentation" of the design.
In this PR the final keyword is used for all local variables in very short methods (e.g. 5-10 LOC), which does not make sense to me.

The objective is to write as little code as possible because that is the easiest to read, maintain.... Look at Java std lib/JDK implementation or read the effective java book.

.github/workflows/build.yaml
```diff
@@ -37,6 +37,12 @@ jobs:
         env:
           JVM_OPTS: -Xmx1g
           TERM: dumb
+
+      - name: Upload javaagent JAR
```
pavolloffay, line 41: this does not seem to belong to this PR

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
```diff
@@ -0,0 +1,105 @@
...
+  public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+    final Map<Junction<MethodDescription>, String> matchers = new HashMap<>();
+    matchers.put(
+        named("read")
+            .and(takesArguments(1))
+            .and(takesArgument(0, ByteBuffer.class))
+            .and(returns(int.class))
+            .and(isPublic()),
+        Read_advice.class.getName());
+    return Collections.unmodifiableMap(matchers);
```
pavolloffay, line 62: There is no need to wrap it into unmodifiable. This code should run as quickly as possible. It literally blocks the app from being started.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
```diff
@@ -0,0 +1,105 @@
...
+public final class StreamSourceChannelInstrumentation implements TypeInstrumentation {
+
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    return AgentElementMatchers.safeHasSuperType(named("org.xnio.channels.StreamSourceChannel"));
+  }
+
+  @Override
+  public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+    final Map<Junction<MethodDescription>, String> matchers = new HashMap<>();
```
pavolloffay, line 54: directly return `singletonMap`

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
```diff
@@ -0,0 +1,105 @@
...
+    matchers.put(
+        named("read")
+            .and(takesArguments(1))
+            .and(takesArgument(0, ByteBuffer.class))
+            .and(returns(int.class))
+            .and(isPublic()),
+        Read_advice.class.getName());
```
pavolloffay, line 61: use string advice name, it's faster than `Class.getName()`, again speed is what we care here. By using a string BB does not have to load the full class.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+    public static void exit(
+        @Advice.Return final int numBytesRead,
+        @Advice.This final StreamSourceChannel streamSourceChannel,
+        @Advice.Thrown final Throwable thrown,
+        @Advice.Argument(0) final ByteBuffer byteBuffer) {
+      if (HypertraceCallDepthThreadLocalMap.decrementCallDepth(StreamSourceChannel.class) > 0
+          || thrown != null) {
```
pavolloffay, line 94: if an exception happened then the counter is not decremented and this advice will never be again executed

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/UndertowHttpServerExchangeInstrumentation.java
```diff
@@ -0,0 +1,83 @@
...
+public final class UndertowHttpServerExchangeInstrumentation implements TypeInstrumentation {
+
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    return failSafe(named("io.undertow.server.HttpServerExchange"));
+  }
+
+  @Override
+  public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+    final Map<Junction<MethodDescription>, String> matchers = new HashMap<>();
```
pavolloffay, line 51: return directly `singletonMap`

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/UndertowInstrumentationModule.java
```diff
@@ -0,0 +1,37 @@
...
+import io.opentelemetry.javaagent.tooling.InstrumentationModule;
+import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
+import java.util.Arrays;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public final class UndertowInstrumentationModule extends InstrumentationModule {
+
+  public UndertowInstrumentationModule() {
+    super("undertow", "undertow-1.4-ht");
```
pavolloffay, line 29: there are missing names `ht` and `undertow-ht`

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
```diff
@@ -0,0 +1,108 @@
...
+        || !ContentTypeUtils.shouldCapture(
+            httpServerExchange.getRequestHeaders().getFirst(Headers.CONTENT_TYPE))) {
+      return;
+    }
+    final Span span =
+        Span.fromContext(UndertowHttpServerTracer.tracer().getServerContext(httpServerExchange));
+    final Charset charset = Charset.forName(httpServerExchange.getRequestCharset());
+    final BoundedByteArrayOutputStream boundedByteArrayOutputStream =
+        BoundedBuffersFactory.createStream(
+            (int) httpServerExchange.getRequestContentLength(), charset);
```
pavolloffay, line 66: what size is used if the content length is `0` e.g. for chunked transfer

instrumentation/undertow-1.4/src/test/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/UndertowInstrumentationTest.java
```diff
@@ -0,0 +1,123 @@
...
+import okhttp3.Response;
+import org.hypertrace.agent.core.instrumentation.HypertraceSemanticAttributes;
+import org.hypertrace.agent.testing.AbstractInstrumenterTest;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+final class UndertowInstrumentationTest extends AbstractInstrumenterTest {
+
+  @Test
+  void testCaptureRequestBody() throws Exception {
```
pavolloffay, line 52: `test` prefix is not used for test methods in this project. No need to write explicit `test` prefix in modern java projects.

nit: `postUrlEncoded` tells me more what the test does.

instrumentation/undertow-1.4/src/test/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/UndertowInstrumentationTest.java
```diff
@@ -0,0 +1,123 @@
...
+import okhttp3.Request;
+import okhttp3.Request.Builder;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import org.hypertrace.agent.core.instrumentation.HypertraceSemanticAttributes;
+import org.hypertrace.agent.testing.AbstractInstrumenterTest;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+final class UndertowInstrumentationTest extends AbstractInstrumenterTest {
```
pavolloffay, line 49: It would be great to write more tests:
* get JSON
* get text - to test no payload is captured

pavolloffay (commented):
Undertow support should be added to the readme.

Should this instrumentation capture headers as well? This is the first servlet container instrumentation in this project. Could the payloads be captured twice? E.g. first in undertow and then in the servlet instrumentation?

instrumentation/undertow-1.4/src/test/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/UndertowInstrumentationTest.java
ryandens: Agreed!

ryandens:
> Undertow support should be added to the readme.

I'm avoiding adding undertow support to the readme because we specifically only will only fully work on Undertow servlets. Due to the implementation details of APIs like `io.undertow.servlet.spec.HttpServletRequestImpl.getParameterMap()`, we know that the request body won't always be accessed via the standard servlet body access APIs `getReader` and `getInputStream()` so we add this as a way to support servlets running on Undertow, but it does not mean that we support running on a raw Undertow HTTP server ( we don't). 

> Should this instrumentation capture headers as well? 

Theoretically, we could add that over time, but the goal of this PR is to solve the request bodies missing when running undertow with Servlets, one of our supported use cases. Right now, adding full-out undertow support isn't our top priority


> This is the first servlet container instrumentation in this project. Could the payloads be captured twice? E.g. first in Undertow and then in the servlet instrumentation?

I'm really glad you asked this! This was indeed happening when you brought this up. The solution I came up with was to add some lightweight instrumentation to `undertow-servlet` to mark and store in the instrumentation context whether or not an `HttpServerExchange` should be instrumented with undertow specific body-capturing.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
ryandens: The `BoundedBuffersFactory` has some logic to help us out with that. In the case of stream being created in this way, we'll simply use the default `ByteArrayOutputStream` constructor which uses the default initial size of 32. Regardless, the ByteArrayOutputStream will grow to fit the size of its content as we write to it (until restricted by the `BoundedByteArrayOutputStream`). I'm sure we can improve on this, but I'm trying to keep this PR scoped to the bug with the customer as I don't have great data on how/when we can improve this right now

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
ryandens: I would love to get some automated benchmarks so we can work to improve our startup time  🙌 so we can know what optimizations matter what ones don't. Soon ⏩ 🤞

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/StreamSourceChannelInstrumentation.java
ryandens: I'm a big fan of effective java! I don't recall an item about final and/or non-final local variables, but I personally find that flipping the switch in the IDE and defaulting to final local variables makes the code easier to maintain and build upon over time. The cost of making variables final is low (because it can be done automatically) and IMO it helps make the code a little more readable because its clear at variable declaration time when a variable is reassigned, rather than relying on the reader to figure out whether a variable is effectively final (because it's never reassigned) or if it's actually mutated. This scales well over time as a function organically grows in length over time and is modified by multiple authors, who may or may not have context on what the previous person wrote. I guess I just disagree that the least amount of code possible is the easiest to read and maintain. A good counterexample to that point, in my opinion, is how books like effective java encourage the use of inheritance over composition. In general, using composition rather than inheritance results in more code (as measured by LOC) but most folks agree that composition is a better default method of sharing code because it is easier to understand what is happening when read in the future. That's not to say that there isn't a place for inheritance (or non-final local variables), just that it's part of what influences what I tend to do by default. 


That being said, in the context of this PR it isn't terribly important, so if you feel strongly about this I'd be happy to table the discussion for another time and remove the `final` keyword from local variables in this PR. We can address it at a time when we can also add in automation to enforce whatever decision we come to.

javaagent-core/src/main/java/org/hypertrace/agent/core/instrumentation/RequestBodyCaptureMethod.java
```diff
@@ -0,0 +1,37 @@
...
+package org.hypertrace.agent.core.instrumentation;
+
+/**
+ * Describes instrumentation strategies for capturing the request body on containers. Generally,
+ * this is detected at runtime and stored in the {@link
+ * io.opentelemetry.javaagent.instrumentation.api.InstrumentationContext} so that {@link
+ * #APP_SERVER} instrumentation can short-circuit when it is know that that the preferred {@link
+ * #SERVLET} instrumentation is sufficient to capture the request body
+ */
+public enum RequestBodyCaptureMethod {
```
pavolloffay, line 26: We literally do not need this type in the `javaagent-core`. `javaagent-core` should contain only types that are used by multiple instrumentations. Which does not seem to be the case here. This type can be moved to undertow module. Also only  `SERVLET` value is used in the codebase.

I think I know your intentions with this type and overall approach, however it won't work as a generic solution across instrumentations. We should design an API that will tell you if payload or headers for a specific request were captured and it should work for both client and server instrumentations (clients have the same problem e.g. jax-rs client implementation is often times backed by apache HTTP client...).  The approach in this PR won't work bc it uses instrumentation context with specific types 3rd party library types e.g. `io.undertow.server.HttpServerExchange`.

You might want to look at https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/3e28b01e42311d755db17d698966d8f3bfaea1e3/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java#L88 to see how to design this using the APIs that are available for all instrumentations.

instrumentation/undertow-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/utils/Utils.java
```diff
@@ -0,0 +1,97 @@
...
+      final ByteBuffer readOnlyBuffer, final int numBytesRead, final SpanAndBuffer spanAndBuffer) {
+    if (numBytesRead <= 0) {
+      return;
+    }
+    readOnlyBuffer.position(readOnlyBuffer.position() - numBytesRead);
+
+    final BoundedByteArrayOutputStream boundedByteArrayOutputStream = spanAndBuffer.byteArrayBuffer;
+    for (int i = 0; i < numBytesRead; i++) {
+      final byte b = readOnlyBuffer.get();
+      boundedByteArrayOutputStream.write(b);
```
pavolloffay, line 94: This does not seem efficient, `ByteBuffer` should have an API to return the allocated array.

ryandens: `ByteBuffer.array()` does exist, but it's not guaranteed to work (depending on the kind of `ByteBuffer` being used. As far as I know, it does not work on `DirectByteBuffer`s, only on heap-based buffers. In the case of undertow, the `ByteBuffer` subclass being used in the test case is a `DirectByteBuffer` so `ByteBuffer.array()` will throw an `UnsupportedOperationException`. 

My first pass at this, after discovering that `ByteBuffer.array()` does not suit our needs, was to do something like:
```java
byte[] dst = new byte[numBytesRead];
readOnlyBuffer.get(dst);
boundedByteArrayOutputStream.write(dst);
```

However, after doing this, I looked at the implementation of `ByteBuffer.get(byte[], int, int)`. It looks like this:

```    
public ByteBuffer get(byte[] dst, int offset, int length) {
  checkBounds(offset, length, dst.length);
  if (length > remaining())
    throw new BufferUnderflowException();
  int end = offset + length;
  for (int i = offset; i < end; i++)
      dst[i] = get();
      return this;
}
```

Note `ByteBuffer.get()` is called `n` times, as it is in this method as well. So, by calling `ByteBuffer.get()` directly here, it is invoked the same number of times as `ByteBuffer.get(byte[])`, except that we get to avoid allocating a duplicate `byte[]` just to have it eligible for GC at the end of this utility method. This could put additional GC pressure on the application and result in degraded performance. I'm definitely open to other options of copying data from the `ByteBuffer`! But as far as I know, this is our best option today.

javaagent-core/src/main/java/org/hypertrace/agent/core/instrumentation/RequestBodyCaptureMethod.java
ryandens: Definitely agreed, I'll move this type into a common library project shared by both `undertow` instrumentation projects. Over time, when we have more use-cases like this one, we can design a more general solution. This was definitely premature 🙂

ryandens:
Hey @pavolloffay, thanks for taking the time to review these changes! Do you have any further comments/concerns about this PR?

instrumentation/undertow/undertow-common/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/common/RequestBodyCaptureMethod.java
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * Signals that the request body should be captured relying on the instrumentation of APIs defined
+   * by {@code javax.servlet} implementations
+   */
+  SERVLET,
+  /**
+   * Signals that request body should be captured relying on the instrumentation of APIs defined
+   * inside the app server's implementation that do not relate to {@code javax.servlet} APIs
+   */
+  APP_SERVER
```
pavolloffay, line 36: I believe this is not being used anywhere

instrumentation/undertow/undertow-1.4/build.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+            "io.opentelemetry.javaagent.tooling.muzzle.collector.MuzzleCodeGenerationPlugin",
+            project(":javaagent-tooling").configurations["instrumentationMuzzle"] + configurations.runtimeClasspath
+    ).configure()
+}
+
+val versions: Map<String, String> by extra
+
+dependencies {
+    implementation("io.opentelemetry.javaagent.instrumentation:opentelemetry-javaagent-undertow-1.4:${versions["opentelemetry_java_agent"]}")
+    library("io.undertow:undertow-core:2.0.0.Final")
```
pavolloffay, line 29: I think this should be 1.4.0. It is what muzzle check is configred against. The idea is to use the lowest version possible.

pavolloffay (commented):
Why there are 3 gradle modules? Any real justification?

There is: 
* common
* undertow
* undertow-servlet 

Does the undertow instrumentation work without servlet? I guess the servlet depends on the core so if the core is not reused I would merge them together.

ryandens:
> Why there are 3 gradle modules? Any real justification?
> 
> There is:
> 
>     * common
> 
>     * undertow
> 
>     * undertow-servlet
> 
> 
> Does the undertow instrumentation work without servlet? I guess the servlet depends on the core so if the core is not reused I would merge them together.

Yes, the common module is depend on by both the undertow module and the undertow-servlet module. The undertow module works without the servlet module because we can't count on undertow-servlet always being used. The undertow-servlet insrumentation module serves to add instrumentation to undertow-servlet APIs in a way that can be detected by our undertow-core instrumentation. This allows us to intelligently avoid capturing the reqeust body twice depeneidng on on what APIs are used to access it/in what context.

instrumentation/undertow/undertow-1.4/build.gradle.kts
ryandens: Works for me 👍  FWIW i copied this part from the [OTEL undertow module](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/undertow-1.4/javaagent/build.gradle#L15)

PR #322, opened 2021-06-17
Move agent config and its dependencies into agent classloader

### LINKED ISSUE #313: GRPC Instrumentation not adding rpc.body to spans
The GRPC instrumentation is not able to add rpc.request.body and rpc.response.body attributes to spans as the protobuf classes are shaded in the javaagent so the instrumentation looks for shaded classes, while during runtime the actual classes are used. 

So we should not shade the protobuf dependencies in javaagent. But javaagent also needs to have protobuf dependency for config. So we need to have the protobuf in javaagent to be in agentClassLoader, and do not shade it so that GRPC instrumentation works.

pavolloffay:

## Description

Resolves #313 

Notable changes:
* config and its dependencies are moved to agent classloarder. The classes in the agent classloader are totally isolated from the application. 
* shading for proto and jackson is removed - not needed anymore
* added a new config interface that is backed by old config class. The new interface is located in `javaagent-core` that lives in the bootstrap classloader. The binding between the interface and proto-based config is done at javaagent bootstrap via component installer SPI.

The whole idea is that `javaagent-core` is located in the bootstrap classloader hence it should not bring any other 3rd party dependencies. The otel`-extensions` is located in the agent classloader and can load and use anything bc it is fully isolated from the application.


pavolloffay:
@ryandens it seems that snyk still fails on the PRs from forks.

pavolloffay:
The CI passes not. Snyk is misconfigured and fails on PRs from forks on 

> "error-message": "`snyk` requires an authenticated account. Please run `snyk auth` and try again.",

pavolloffay:
The body capture works, tested on https://github.com/grpc/grpc-java/tree/master/examples 

Screenshot of Jaeger UI

pavolloffay:
do not merge yet it needs more love - e.g. remove the filter paths from the instrumentation config. That should not be available to instrumentations.

instrumentation/netty/netty-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/netty/v4_0/server/HttpServerRequestTracingHandler.java
```diff
@@ -86,7 +86,7 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
     }
 
     if ((msg instanceof HttpContent || msg instanceof ByteBuf)
-        && agentConfig.getDataCapture().getHttpBody().getRequest().getValue()) {
+        && instrumentationConfig.rpcBody().request()) {
```
shashank11p, line 89: Should this be httpBody?

javaagent-core/src/main/java/org/hypertrace/agent/core/config/InstrumentationConfig.java
```diff
@@ -0,0 +1,91 @@
...
+    boolean request();
+
+    boolean response();
+  }
+
+  default boolean isInstrumentationEnabled(String primaryName, String[] otherNames) {
+    // the instNames is not used because the config does not support it at the moment.
+
+    // disabled if all is disabled
+    if (!httpBody().request()
```
shashank11p, line 48: Dont we want to check rpcBody?

pavolloffay: good catch!

pavolloffay:
@shashank11p PR updated, could you please re-review?

ryandens:
> @ryandens it seems that snyk still fails on the PRs from forks.

Hey Pavol! Sorry about that would you mind trying out a fix to help this PR along? I think we need to 

1. Swap out `--org=hypertrace` with `--org=$GITHUB_ACTOR` 
2. Depending on whether you've done this before, you'll need to login and authorize your github account.

pavolloffay:
I am not sure if that will work, most likely the secret is not present in PRs from forks https://github.com/pavolloffay/javaagent/blob/52f0a2d388ce5951e4627f6f18c0fa0f9e4f12a1/.github/workflows/build.yaml#L82.

shashank11p (approved):

Issue #238, opened 2021-01-18
Run tests with javaagent


pavolloffay:
This separately tracks https://github.com/hypertrace/javaagent/issues/234. 

Switch to https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1643 to run unit tests in instrumentation modules as we are still using test running approach from 0.13.x.

- [ ] remove testing-bootstrap module

PR #323, opened 2021-06-21
Move mock filter impl to testing package

pavolloffay:

## Description

Before this PR the mock filter implementation was packaged in the agent distribution. The mock filter shouldn't be exposed in the final distribution. Its only purpose is to test blocking in instrumentations which is done by moving the filter into the testing package.


smoke-tests/src/test/java/org/hypertrace/agent/smoketest/SpringBootSmokeTest.java
```diff
@@ -148,20 +148,4 @@ public void get() throws IOException, InterruptedException {
     //    Assertions.assertEquals(1, responseBodyAttributes.size());
     //    Assertions.assertEquals("Hi!", responseBodyAttributes.get(0));
   }
-
-  @Test
-  public void blocking() throws IOException {
```
pavolloffay, line 153: this is obviously a loss. To make this test work with mock filer not being in the final distribution we will have to build test java agent distro - this might be related to https://github.com/hypertrace/javaagent/issues/238

shashank11p (approved):

PR #325, opened 2021-06-21
Fix failing snyk running on forks

pavolloffay:


pavolloffay:
The issue is that the `SNYK_TOKEN` is not added to the build. It worked in the ht-service https://github.com/hypertrace/hypertrace-service/runs/2876248502#step:4:5

pavolloffay:
The solution is to use `pull_request_target` conf in the workflow config e.g. https://github.com/hypertrace/hypertrace-service/blob/main/.github/workflows/pr-build.yml#L6. 

This makes the secrets available to the fork builds https://docs.github.com/en/actions/reference/events-that-trigger-workflows

pavolloffay:
@ryandens see now https://github.com/hypertrace/javaagent/pull/323 is passing.

PR #331, opened 2021-07-06
Fix flaky undertow test

pavolloffay:


instrumentation/undertow/undertow-1.4/src/test/java/io/opentelemetry/javaagent/instrumentation/hypertrace/undertow/v1_4/UndertowInstrumentationTest.java
```diff
@@ -108,8 +108,8 @@ void postUrlEncoded() throws InterruptedException, TimeoutException, IOException
       assertEquals(200, response.code());
     }
 
-    List<List<SpanData>> traces = TEST_WRITER.getTraces();
```
pavolloffay, line 111: @ryandens another comment on using `final` everywhere. The test class looks messy, I am obviously against using `final` where no necessary, but here it's even messier bc it's not used consistently. It degrades code quality.

pavolloffay:
It seems that the fix didn't work, the first build failed on the same error.

```
UndertowInstrumentationTest > postUrlEncoded() FAILED
    org.opentest4j.AssertionFailedError at UndertowInstrumentationTest.java:113
```

pavolloffay:
I will merge this as it is an improvement, the traces should be queried after wait statement.

PR #336, opened 2021-07-13
🐛  fix bug where release pipeline is triggered on every branch creation

ryandens:
## Description
Currently, when a new branch is created, a release pipeline is triggered. i believe that is because the syntax of the GitHub actions workflow for our release pipeline needs to be tweaked to use the YAML path `on.push.tags` rather than `on.create.tags`. 

See:
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags
- https://github.community/t/how-to-run-github-actions-workflow-only-for-new-tags/16075/18

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,7 +1,7 @@
 name: release
 on:
-  create:
-    tags: ['^[0-9]+\.[0-9]+\.[0-9]+']
+  push:
+    tags: ['[0-9]+\.[0-9]+\.[0-9]+']
 
 jobs:
   test:


ryandens:
I noticed that this occurred when I created a new branch, so I cancelled the workflow. This has been happening for a while, but up until the migration away from CircleCI this pipeline was essentially a no-op so it did no harm. Now, it could

ryandens:
Example canceled workflow: https://github.com/hypertrace/javaagent/actions/runs/1026610635

davexroth (approved):

pavolloffay:
It's a common practice to use branches in forks and keep the upstream with only main or maintenance branches.

pavolloffay:
I am curious do you have a link to CI run for your branch? The release job uses pattern match (e.g. `x.x.x`) that should not match your feature branch.

ryandens:
The problem was the syntax of the yaml file, which was  the yaml path `on.create.tags`. This isn't a valid configuration option for GitHub actions, it should be `on.push.tags` as this PR changes it to be. Now, the pattern match is enforced

PR #345, opened 2021-08-27
Upgrade otel 1 4 1

ryandens:
## Description

### Testing

### Checklist:

### Documentation


ryandens:
This PR obviates the need for #341 and helps us avoid the un-shading 😄

smoke-tests/src/test/resources/otel.yaml
```diff
@@ -27,7 +27,7 @@ exporters:
 service:
   pipelines:
     traces:
-      receivers: [otlp]
+      receivers: [otlp, zipkin, jaeger]
```
pavolloffay: since we have never supported jaeger we should just remove it from here.

buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
```diff
@@ -43,6 +51,40 @@ public void apply(Project project) {
             });
   }
 
+  private void addDependencies(Project project) {
```
pavolloffay, line 54: Am I right that all projects that use auto-instrumenation-plugin will add the following dependencies?

instrumentation/apache-httpasyncclient-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/apachehttpasyncclient/ApacheAsyncClientInstrumentationModule.java
```diff
@@ -95,70 +103,113 @@ public int order() {
...
+    private static final Logger log =
+        LoggerFactory.getLogger(DelegatingCaptureBodyRequestProducer.class);
+
+    private static final Lookup lookup = MethodHandles.lookup();
+    private static final MethodHandle getContext;
+
+    static {
+      MethodHandle localGetContext;
+      try {
+        Field contextField = WrappedFutureCallback.class.getDeclaredField("context");
```
pavolloffay, line 137: Could you please create an issue in the OTEL upstream to expose the context and add TODO comment here with the link.

instrumentation/build.gradle.kts
```diff
@@ -44,6 +18,7 @@ dependencies{
     implementation(project(":instrumentation:netty:netty-4.1"))
     implementation(project(":instrumentation:undertow:undertow-1.4"))
     implementation(project(":instrumentation:undertow:undertow-servlet-1.4"))
+    implementation(project(":instrumentation:vertx:vertx-web-3.0"))
```
pavolloffay, line 21: good catch :)

instrumentation/vertx/vertx-web-3.0/build.gradle.kts
```diff
@@ -18,6 +18,7 @@ muzzle {
         group = "io.vertx"
         module = "vertx-web"
         versions = "[3.0.0,4.0.0)"
+//        extraDependency("io.netty:netty-codec-http:4.0.38.Final")
```
pavolloffay, line 21: remove?

testing-common/src/testFixtures/java/org/hypertrace/agent/testing/TestOpenTelemetryInstaller.java
```diff
@@ -33,13 +34,25 @@ public TestOpenTelemetryInstaller(SpanProcessor spanProcessor) {
...
   @Override
-  public void beforeByteBuddyAgent(Config config) {
+  public void beforeAgent(Config config) {
     OpenTelemetrySdk.builder()
         .setTracerProvider(SdkTracerProvider.builder().addSpanProcessor(spanProcessor).build())
         .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
         .buildAndRegisterGlobal();
+
+    try {
+      Class<?> agentInitializerClass =
```
pavolloffay, line 44: Could you please document this work-around? Is there upstream issue to handle this properly?

pavolloffay (commented):
Great work 💯 . This seems much better than the shadowing approach in the previous PR. 

Added some minor comments.

buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
ryandens: Correct!

instrumentation/apache-httpasyncclient-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/apachehttpasyncclient/ApacheAsyncClientInstrumentationModule.java
ryandens: Unfortunately, I'm not sure that it's a reasonable expectation for them to do this. Before, we accessed the context via a public access to a `final` field on `DelegatingRequestProducer`. However, in [OTEL java-instrumentation pull 3209](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3209/files) a change was made to postponing http client span creation until the request is available. This meant that the `DelegatingRequestProducer` no longer has a client context when it is instantiated, but rather sets a new `volatile` field on `WrappedFutureCallback`. I don't think it's a reasonable ask to expose this volatile field, and I can see why they would not want to expose a nullable volatile field to downstream instrumentation libraries as it can only be invoked after the `DelegatingRequestProducer.generateRequest`. IMO I think we should refactor our implementation of this instrumentation to not rely on this field and instead be able to do something like `Java8BytecodeBridge.currentSpan()` but for some reason I can't get that to work properly. I'll add a comment indicating this

testing-common/src/testFixtures/java/org/hypertrace/agent/testing/TestOpenTelemetryInstaller.java
ryandens: No upstream issue here because I'm working on a strategy to get us leveraging more of their test fixtures

instrumentation/netty/netty-4.1/build.gradle.kts
```diff
@@ -42,8 +42,8 @@ afterEvaluate{
 val versions: Map<String, String> by extra
 
 dependencies {
+    implementation("io.opentelemetry.instrumentation:opentelemetry-netty-4.1:${versions["opentelemetry_java_agent"]}")
     implementation("io.opentelemetry.javaagent.instrumentation:opentelemetry-javaagent-netty-4.1:${versions["opentelemetry_java_agent"]}")
```
shashank11p, line 46: In other frameworks above, we have made this testImplementation.

instrumentation/netty/netty-4.1/build.gradle.kts
```diff
@@ -42,8 +42,8 @@ afterEvaluate{
 val versions: Map<String, String> by extra
 
 dependencies {
+    implementation("io.opentelemetry.instrumentation:opentelemetry-netty-4.1:${versions["opentelemetry_java_agent"]}")
```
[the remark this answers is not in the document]
ryandens: `netty-4.0` does not have a "standalone library" instrumentation module like `netty-4.1` does. netty-4.0 depends on the `javaagent` module to get compile-time access in our deployed code to code defined in OTEL (like tracers, attribute keys, etc)

instrumentation/netty/netty-4.1/build.gradle.kts
ryandens: We can move it to `testImplementation` when our code in `src/main` does not reference OTEL code defined in the instrumentation module. Ideally, we would only have to depend on `io.opentelemetry.instrumentation:opentelemetry-netty-4.1` for compiling our main source set and use the  `io.opentelemetry.javaagent.instrumentation:opentelemetry-javaagent-netty-4.1` in our test runtime. However, we need compile access to things like the `NettyHttpServerTracer` in our main source set

davexroth (approved):

PR #349, opened 2021-11-04
✨ Evaluate http bodies for Java EE with Hypertrace Filter

ryandens:
### Testing

### Checklist:

### Documentation

### DIFF
--- a/instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/servlet/v3_0/nowrapping/Servlet30AndFilterInstrumentation.java
+++ b/instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/servlet/v3_0/nowrapping/Servlet30AndFilterInstrumentation.java
@@ -46,6 +46,7 @@ import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.matcher.ElementMatcher;
 import org.hypertrace.agent.core.config.InstrumentationConfig;
 import org.hypertrace.agent.core.instrumentation.HypertraceCallDepthThreadLocalMap;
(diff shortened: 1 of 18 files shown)


javaagent-core/src/main/java/org/hypertrace/agent/core/instrumentation/buffer/ByteBufferSpanPair.java
```diff
@@ -44,5 +50,10 @@ public void captureBody(AttributeKey<String> attributeKey) {
       // ignore charset has been parsed before
     }
     span.setAttribute(attributeKey, requestBody);
+    final boolean block;
+    block = filter.apply(span, requestBody);
```
ryandens, line 54: make sure  to update this to provide headers to this filter

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/servlet/v3_0/nowrapping/Servlet30AndFilterInstrumentation.java
```diff
@@ -130,13 +124,24 @@ public static boolean start(
         // skip execution of the user code
         return true;
       }
+
```
ryandens, line 127: We defer putting these in the instrumentation context until after `evaluateRequestHeaders` is called for two reasons:
1. If `evaluateRequestHeaders` returns `true`, these objects don't need to be in the instrumentation context because the request body will never be read
2. The header map needs to be passed along with the `Span` now so that the call to `evaluateRequestBody` can have access to the HTTP headers to make intelligent choices about whether the request should be blocked based on the value of the headers and the body, not just the body.

shashank11p (approved):

PR #350, opened 2021-11-10
Add metric config and send metrics

shashank11p:
Update agent-config to have metrics configurations, and update javaagent to use them and send metrics using corresponding exporter from config.

Default metric exporter type to OTLP.


otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/EnvironmentConfig.java
```diff
@@ -130,6 +133,14 @@ private static void applyPropagationFormat(AgentConfig.Builder builder) {
     if (traceReportingType != null) {
       builder.setTraceReporterType(TraceReporterType.valueOf(traceReportingType));
     }
+    String metricReporterAddress = getProperty(REPORTING_METRIC_ENDPOINT);
```
ryandens, line 136: should these have default values? With reporting traces, we default to localhost:4317 and OTLP, meaning that for most use-cases users only have to configure the metric endpoint and not the metric type and for simple use cases no configuration is required at all

otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/EnvironmentConfig.java
```diff
@@ -130,6 +133,14 @@ private static void applyPropagationFormat(AgentConfig.Builder builder) {
...
+    if (metricReporterAddress != null) {
```
ryandens, line 137: The `EnvironmentConfig` test should be expanded to also test these new properties

otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/HypertraceConfig.java
```diff
@@ -143,9 +144,18 @@ static AgentConfig load(String filename) throws IOException {
     if (!builder.hasEndpoint()) {
       builder.setEndpoint(StringValue.newBuilder().setValue(DEFAULT_REPORTING_ENDPOINT).build());
     }
-    if (builder.getTraceReporterType().equals(Config.TraceReporterType.UNSPECIFIED)) {
+    if (!builder.hasMetricEndpoint()) {
```
ryandens, line 147: The test cases in `HypertraceConfigTest` should be expanded to include testing for these new properties

ryandens (changes requested):
🎉 Really excited to get metrics in here! Would it be possible to extend our `SpringBootSmokeTest` to also verify we report metrics as expected?

otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/EnvironmentConfig.java
davexroth: we should default the metrics endpoint to the whatever the traces endpoint is.

shashank11p: Yes, these have default values and it is the same as traces. The default type is OTLP, and endpoint is also same as traces endpoint.

otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/EnvironmentConfig.java
shashank11p: done

otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/HypertraceConfig.java
shashank11p: done

otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/HypertraceConfig.java
```diff
@@ -143,9 +144,18 @@ static AgentConfig load(String filename) throws IOException {
...
+      builder.setMetricEndpoint(
+          StringValue.newBuilder().setValue(DEFAULT_REPORTING_ENDPOINT).build());
```
ryandens, line 149: I believe what we want here is to set the value of the metric endpoint to the value of the trace reporting endpoint, rather than having them default to the same value. That way, to get traces and metrics, most users will only have to configure the reporting endpoint. However, if they want to report to a different endpoint (e.g. to export Prometheus metrics) they can override it. I think this should be:
```
if (!builder.hasMetricEndpoint()) {
  builder.setMetricEndpoint(builder.getEndpoint());
}
```
At this point, the builder.getEndpoint is guaranteed to be set (either to the default value or one interpreted from the environment

shashank11p: Yeah if we have a single endpoint for both traces and metrics, that would be better. But it will need to be a breaking change. Because as it is now, the endpoint is only for traces, for example, in case of ZIPKIN trace reporter type, the endpoint would be something like `http://localhost:9411/api/v2/spans` which will not work for metrics if the metrics config is not given. 
The best way would be to have a generic endpoint and two other endpoints for traces and metrics if need to override. But since we already are using endpoint for traces, it will be a breaking change.

shashank11p: Updated to use reporter endpoint if metrics endpoint is not set

ryandens:
From a code perspective, this is looking great! My one last question is regarding backward compatibility. For users still using the zipkin exporter type, the metric exporter will default to `localhost:4317`. However, `localhost:4317` may not be available/reachable, and could place an unnecessary burden on the application JVM. Should we add a check to not export metrics if the Zipkin endpoint is set?

shashank11p:
> From a code perspective, this is looking great! My one last question is regarding backward compatibility. For users still using the zipkin exporter type, the metric exporter will default to `localhost:4317`. However, `localhost:4317` may not be available/reachable, and could place an unnecessary burden on the application JVM. Should we add a check to not export metrics if the Zipkin endpoint is set?

That is a good point. So we would be checking trace reporter type for disabling metrics. The problem in checking if trace reporter type is ZIPKIN or not is that metric reporter type can be OTLP with metrics endpoint reachable even when traces endpoint is zipkin. So we will disable metrics when they should not have been disabled.

One way I can think of is to add another metricReportingType as `none` in agent-config and if someone is using ZIPKIN, they can set this config for metricReporterType to `none`. And then we will disable exporting metrics in agent.

ryandens:
I think my take would be:

1. If traceReporting type is zipkin and metric reporting type/endpoint is not set (e.g. default) we shouldn't start reporting metrics today and we should configure the SDK to not report  metrics (current behavior)
2. If traceReporting type is zipkin and metric reporting type is set (but endpoint is not set) then we should default the metric endpoint to `localhsot:4317` for otlp metrics or `localhost:9464` for prometheus 
3. If traceReporting type is zipkin and metric reporting endpoint is set (but metric type is not set) then we should default the metric type to otlp

ryandens:
Good point, I forgot about that quirk. The order in which those tasks are done is definitely problematic and makes a derived configuration hard to work with. This is definitely going to be something we should rework as the agent inevitably builds in complexity, but I'm not sure if tackling that now makes sense. I totally agree we should have a configuration option to set metrics exporter to none.

@davexroth any thoughts here? It seems like our choices are a non-trivial refactor to  better align with OTEL's configuration  or simply defaulting metrics to `oltp` and `localhost:4317` and encouraging users to override as appropriate

davexroth:
> I think my take would be:
> 
> 1. If traceReporting type is zipkin and metric reporting type/endpoint is not set (e.g. default) we shouldn't start reporting metrics today and we should configure the SDK to not report  metrics (current behavior)
> 2. If traceReporting type is zipkin and metric reporting type is set (but endpoint is not set) then we should default the metric endpoint to `localhsot:4317` for otlp metrics or `localhost:9464` for prometheus
> 3. If traceReporting type is zipkin and metric reporting endpoint is set (but metric type is not set) then we should default the metric type to otlp

I think this is close.  For 2. if a user sets metrics reporting type is set, but endpoint isn't, that's an error case, and we should just disable metrics.

ryandens:
Totally agreed, but Shashank points out some problems to implementing that in our current context (as a result of how config is applied). Reworking this is possible, but complex and shouldn't be done lightly without also thinking about a bigger picture perspective of how we want config to work long term IMO.

davexroth:
> Totally agreed, but Shashank points out some problems to implementing that in our current context (as a result of how config is applied). Reworking this is possible, but complex and shouldn't be done lightly without also thinking about a bigger picture perspective of how we want config to work long term IMO.

If it's too big a change, I'm ok with just saying that we turn off metrics when zipkin trace exporter is selected - it's not something we ever recommend to use these days.

shashank11p:
> If it's too big a change, I'm ok with just saying that we turn off metrics when zipkin trace exporter is selected - it's not something we ever recommend to use these days.

Okay, so if zipkin trace exporter is selected we will not send metrics. 
And I think we should also add a `none` reporter type in metric to disable it if the collector to which data is being sent only has traces pipeline but not metrics pipeline. 

So we will not send metrics when ZIPKIN trace reporter is selected or when metricReportingType is set to none. Is this good?

ryandens:
> So we will not send metrics when ZIPKIN trace reporter is selected or when metricReportingType is set to none. Is this good?

This works for me - though FWIW I think we don't need to add a "none" type to the metric exporter. We can simply not set a default and treat the `METRIC_REPORTER_TYPE_UNSPECIFIED` value as "none".  If possible, we should add some more test cases to cover the corner cases discussed in this PR to help prevent regressions

ryandens:
I was struggling to reason with what was working/what wasn't, so I wound up writing a few test cases to explore the zipkin trace exporter edge cases:

zipkinConfig.yaml
```yaml
service_name: service
reporting:
  trace_reporter_type: ZIPKIN
  endpoint: http://example.com:9411/api/v2/spans
```
```
@Test
  @SetEnvironmentVariable(key = "HT_REPORTING_ENDPOINT", value = "http://oltp.hypertrace.org:4317")
  public void complexConfig() throws IOException {
    // GIVEN a config file with a non-default reporting endpoint and an env-var with a different non-default reporting endpoint
    URL resource = getClass().getClassLoader().getResource("config.yaml");
    // WHEN we load the config
    AgentConfig agentConfig = HypertraceConfig.load(resource.getPath());
    // VERIFY the trace and metric endpoints are the both the value of the env var
    String expectedEndpoint = "http://oltp.hypertrace.org:4317";
    assertEquals(expectedEndpoint, agentConfig.getReporting().getEndpoint().getValue());
    assertEquals(expectedEndpoint, agentConfig.getReporting().getMetricEndpoint().getValue());
  }

  @Test
  public void zipkinExporter() throws IOException {
    // GIVEN a config file with a non-default zipkin reporting endpoint
    URL resource = getClass().getClassLoader().getResource("zipkinConfig.yaml");
    // WHEN we load the config
    AgentConfig agentConfig = HypertraceConfig.load(resource.getPath());
    // VERIFY the trace reporting endpoint is the zipkin endpoint
    assertEquals("http://example.com:9411/api/v2/spans", agentConfig.getReporting().getEndpoint().getValue());
    // VERIFY the metric endpoint is not set
    assertEquals( "", agentConfig.getReporting().getMetricEndpoint().getValue());
  }

  @Test
  @SetEnvironmentVariable(key = "HT_REPORTING_METRIC_REPORTER_TYPE", value = "METRIC_REPORTER_TYPE_OTLP")
  public void zipkinExporterOtlpMetricsType() throws IOException {
    // GIVEN a config file with a non-default zipkin reporting endpoint and the metric type is set in an env var
    URL resource = getClass().getClassLoader().getResource("zipkinConfig.yaml");
    // WHEN we load the config
    AgentConfig agentConfig = HypertraceConfig.load(resource.getPath());
    // VERIFY the trace reporting endpoint is the zipkin endpoint
    assertEquals("http://example.com:9411/api/v2/spans", agentConfig.getReporting().getEndpoint().getValue());
    // VERIFY the metric reporting type is otlp
    assertEquals(MetricReporterType.METRIC_REPORTER_TYPE_OTLP, agentConfig.getReporting().getMetricReporterType());
    // VERIFY the metric endpoint is not set
    assertEquals( "", agentConfig.getReporting().getMetricEndpoint().getValue());
  }

  @Test
  @SetEnvironmentVariable(key = "HT_REPORTING_METRIC_ENDPOINT", value = "http://oltp.hypertrace.org:4317")
  public void zipkinTraceExporterOtlpMetricsEndpoint() throws IOException {
    URL resource = getClass().getClassLoader().getResource("zipkinConfig.yaml");
    AgentConfig agentConfig = HypertraceConfig.load(resource.getPath());
    String expectedEndpoint = "http://example.com:9411/api/v2/spans";
    assertEquals(expectedEndpoint, agentConfig.getReporting().getEndpoint().getValue());
    assertEquals( "http://oltp.hypertrace.org:4317", agentConfig.getReporting().getMetricEndpoint().getValue());
    assertEquals(MetricReporterType.METRIC_REPORTER_TYPE_OTLP, agentConfig.getReporting().getMetricReporterType());
  }
```

Based on the decisions in this PR, i don't know how useful they are but figured they were worth preserving in some fashion

shashank11p:
As discussed offline, we will disable metric reporting if TRACE_REPORTER_TYPE is ZIPKIN, or if METRIC_REPORTER_TYPE is set to `none`.

ryandens (approved):

otel-extensions/src/test/java/org/hypertrace/agent/otel/extensions/config/EnvironmentConfigTest.java
```diff
@@ -90,5 +100,11 @@ public void systemProperties() {
     Assertions.assertEquals(
         StringValue.newBuilder().setValue("/path/2/jar.jar").build(),
         agentConfig.getJavaagent().getFilterJarPaths(1));
+
+    // Assert that metrics endpoint is added correctly if set
+    System.setProperty(EnvironmentConfig.REPORTING_METRIC_ENDPOINT, "https://:-)");
```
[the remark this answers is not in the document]
shashank11p: I used this here and again calling `applyPropertiesAndEnvVars` because in assertions above it is testing without this property set. Then here it is updating the systemProperties and then testing. I changed `http` to `https` to be sure it is taking the systemProperty set now. 

Same for assertion below this.

otel-extensions/src/test/resources/config.yaml
```diff
@@ -5,8 +5,10 @@ propagationFormats:
   - B3
 reporting:
   endpoint: http://localhost:4317
+  metric_endpoint: http://localhost:4317
   secure: true
   trace_reporter_type: OTLP
+  metric_reporter_type: METRIC_REPORTER_TYPE_OTLP
```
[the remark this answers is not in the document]
shashank11p: That is an issue. As protos dont allow two enums with same name in one package. OTLP is used in TraceReporterType. So we cannot reuse that in MetricReporterType as they are in same package. So enum names in protos are very often prefixed with type like we have here.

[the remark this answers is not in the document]
ryandens: weird, this comment is an old one of mine but for some reason github says i added it two hours ago

otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/HypertraceConfig.java
```diff
@@ -143,9 +144,18 @@ static AgentConfig load(String filename) throws IOException {
     if (!builder.hasEndpoint()) {
       builder.setEndpoint(StringValue.newBuilder().setValue(DEFAULT_REPORTING_ENDPOINT).build());
     }
-    if (builder.getTraceReporterType().equals(Config.TraceReporterType.UNSPECIFIED)) {
+    if (!builder.hasMetricEndpoint()) {
+      // If metric endpoint is not given, use the reporter endpoint
+      builder.setMetricEndpoint(builder.getEndpoint());
```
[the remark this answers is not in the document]
shashank11p: Thanks. I had missed this. Updated now.

ryandens (approved):

PR #352, opened 2021-11-20
fix vulnerabilities

shashank11p:


ryandens:
Agreed this is a good quick fix, let's also follow up with upgrades to the parent protobuf and gson libraries

ryandens (approved):

PR #357, opened 2022-01-21
Upgrade otel 1 10

ryandens:
## Description
Upgrades OpenTelemetry to 1.10.0

- updates the instrumentation unit test harness to be compatible with changes to the upstream tooling. Notably, we had to add some workarounds to make sure extensions get loaded for the system class loader in the scope of unit tests
- add `opentelemetry-javaagent-servlet-common-bootstrap` dependencies to test runtime of instrumentation modules that required it (those that use the servlet instrumentation module). 
- Make muzzle checks compatible with framework-specific bootstrap modules. These are special because in production OpenTelemetry injects these classes so they can be loaded by multiple class loaders (e.g. the application class loader and the agent class loader). Without some help, muzzle can't find these classes while doing checks

### Testing

### Checklist:

### Documentation


buildSrc/src/main/groovy/MuzzlePlugin.groovy
```diff
@@ -80,21 +89,18 @@ class MuzzlePlugin implements Plugin<Project> {
           project.getLogger().info('No muzzle pass directives configured. Asserting pass against instrumentation compile-time dependencies')
           ClassLoader userCL = createCompileDepsClassLoader(project, bootstrapProject)
           ClassLoader instrumentationCL = createInstrumentationClassloader(project, toolingProject)
-          Method assertionMethod = instrumentationCL.loadClass('io.opentelemetry.javaagent.muzzle.matcher.MuzzleGradlePluginUtil')
-            .getMethod('assertInstrumentationMuzzled', ClassLoader.class, ClassLoader.class, boolean.class)
-          assertionMethod.invoke(null, instrumentationCL, userCL, true)
+          MuzzleGradlePluginUtil.@Companion.assertInstrumentationMuzzled(instrumentationCL, userCL, true)
```
ryandens, line 92: this was a bit of a weird change, but we used to use reflection to access this because it wasn't available at plugin compile time. Now it is, and in addition it was refactored to be a kotlin companion object rather than just a java static method. This refactor causes this reflection to fail, so I updated it to just access it normally.

smoke-tests/src/test/java/org/hypertrace/agent/smoketest/SpringBootSmokeTest.java
```diff
@@ -174,7 +174,7 @@ public void postJson() throws IOException, InterruptedException {
 
     ArrayList<ExportMetricsServiceRequest> metrics = new ArrayList<>(waitForMetrics());
     Assertions.assertTrue(hasMetricNamed("otlp.exporter.seen", metrics));
-    Assertions.assertTrue(hasMetricNamed("otlp.exported.exported", metrics));
+    Assertions.assertTrue(hasMetricNamed("otlp.exporter.exported", metrics));
```
ryandens, line 177: looks like this metric name changed to be more consistent

testing-common/src/testFixtures/java/org/hypertrace/agent/testing/TestAgentStarter.java
```diff
@@ -0,0 +1,35 @@
...
+  public boolean delayStart() {
+    return false;
+  }
+
+  @Override
+  public void start() {}
+
+  @Override
+  public ClassLoader getExtensionClassLoader() {
+    return ClassLoader.getSystemClassLoader();
```
ryandens, line 33: we do this to instruct the agent to load extensions from the system class loader - we won't need to do this once we migrate to their build plugins

shashank11p (approved):

PR #364, opened 2022-06-08
ENG-17140  Ensure requests can be filtered because of method body

bradAtTraceable:
## Description

### Testing

### Checklist:

### Documentation


bradAtTraceable:
The purpose of this test is to ensure that the Filter receives the contents of an HTTP request body, so that it can block the request, if required.  Previously, the request body was only being passed to the filter if the reader of the request body received an indication that the end-of-stream was reached.  However, in some cases, the reader would stop reading the request body without seeing the normal end-of-stream indication, and the filter would not be called.

shashank11p (approved):

shashank11p:
@bradAtTraceable Some smoke tests are failing. Please update the tests as well :)

PR #367, opened 2022-08-15
Support configurable Allowed Content Types within Hypertrace Javaagent

bradAtTraceable:
I had to start with a new PR and branch on this issue, due to the problems I was having pushing to the origin branch.  This, I believe, was due to submodules within the Hypertrace agent.  I think that these problems are fixed with this PR.

### DIFF
--- a/otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/DataCaptureConfigImpl.java
+++ b/otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/DataCaptureConfigImpl.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright The Hypertrace Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.hypertrace.agent.otel.extensions.config;
+
+import com.google.auto.service.AutoService;
+import com.google.protobuf.StringValue;
+import java.util.List;
+import org.hypertrace.agent.config.v1.Config;
+import org.hypertrace.agent.core.config.DataCaptureConfig;
+
+/**
+ * An implementation of the DataCaptureConfig interface, which can provide the names of the content
+ * types that can be captured.
+ */
+@AutoService(DataCaptureConfig.class)
+public class DataCaptureConfigImpl implements DataCaptureConfig {
+
+  private final Config.AgentConfig agentConfig = HypertraceConfig.get();
+
+  private final Config.DataCapture dataCaptureConfig =
+      (agentConfig != null) ? agentConfig.getDataCapture() : null;
+
+  private String[] allAllowedContentTypes;
+
+  public DataCaptureConfigImpl() {
+    super();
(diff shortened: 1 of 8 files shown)


otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/HypertraceConfig.java
```diff
@@ -189,4 +219,27 @@ private static String convertYamlToJson(InputStream yaml) throws IOException {
     ObjectMapper jsonWriter = new ObjectMapper();
     return jsonWriter.writeValueAsString(obj);
   }
+
+  /**
+   * Creates a collection of objects consisting of the set of objects specified in the
```
tim-mwangi, line 224: Should we have the config allowed content types just do the override? i.e if you set the config yourself, you need to add the defaults as well in the config i.e

otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/DataCaptureConfigImpl.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.protobuf.StringValue;
+import java.util.List;
+import org.hypertrace.agent.config.v1.Config;
+import org.hypertrace.agent.core.config.DataCaptureConfig;
+
+/**
+ * An implementation of the DataCaptureConfig interface, which can provide the names of the content
+ * types that can be captured.
+ */
+@AutoService(DataCaptureConfig.class)
```
tim-mwangi, line 29: Why can't the default allowed content types be just a list of strings?

otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/HypertraceConfig.java
ryanericson: yep

bradAtTraceable: ok, I wasn't sure what was the right thing to do here.  Will fix

otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/config/DataCaptureConfigImpl.java
bradAtTraceable: I followed the precedent of defining implementations for the portions of the configuration that are required by the Java agent.  The Java agent uses Java services to perform the initialization of config data; I merely continued the practice, for consistency.

tim-mwangi: I see. Ok.

otel-extensions/src/test/resources/nonDefaultDataCaptureConfig.yaml
```diff
@@ -0,0 +1,23 @@
...
+  trace_reporter_type: OTLP
+  metric_reporter_type: METRIC_REPORTER_TYPE_OTLP
+  cert_file: /foo/bar/example.pem
+dataCapture:
+  bodyMaxSizeBytes: 16
+  httpHeaders:
+    request: true
+    response: false
+  httpBody:
+  allowed_content_types: ["foo", "bar"]
```
shashank11p, line 19: Should we only use the ones given here or add these along with the default? I guess using only the ones given in config would give us options to remove some content-types if we do not want to capture those.

tim-mwangi: https://github.com/hypertrace/javaagent/pull/367#discussion_r946068547

tim-mwangi: We'd like the config to override the defaults. So if you set the config then only those in the config will be used.

tim-mwangi (approved):

PR #371, opened 2022-09-09
        Create 'otel.extension.library.version' attribute in all spans

bradAtTraceable:
Because the otel.library.version span attribute contains the OTEL version, rather than our version, this PR is introducing a new attribute: 'otel.extension.library.version', which will contain the version of the Hypertrace agent.


shashank11p (commented):
We can maybe add this in https://github.com/hypertrace/javaagent/blob/main/otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/processor/AddTagsSpanProcessor.java#L39, so that we don't have to add it in every instrumentation framework, just at one place, and for any new frameworks in the future we do not need to remember and take care of adding this.

PR #373, opened 2022-10-20
Catch our blocking exception when it is nested.  Eg. spring framework.

tyroneyu:
## Description

When a filter blocks a request (eg. traceable filter), the http code returned is 500. It should be 403. The happens when the exception that we throw is nested and we are not catching it properly. This can be produced when used the spring framework as an app.

### Testing
Manual tested with a spring framework app

### Checklist:

### DIFF
--- a/instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/servlet/v3_0/nowrapping/Servlet30AndFilterInstrumentation.java
+++ b/instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/servlet/v3_0/nowrapping/Servlet30AndFilterInstrumentation.java
@@ -205,12 +205,16 @@ public class Servlet30AndFilterInstrumentation implements TypeInstrumentation {
           }
         }
       } finally {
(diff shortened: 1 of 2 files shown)


shashank11p (approved):
Nice! While we are at it, can we add this loop at other places where we are catching this exception as well?

tyroneyu:
> Nice! While we are at it, can we add this loop at other places where we are catching this exception as well?

I only see one spot where we "catch" this exception.  The other areas just check and rethrow the exception. In those try blocks, they all call `ByeBufferSpanPair.captureBody` and that is where the exception is thrown.

shashank11p (approved):

shashank11p:
> > Nice! While we are at it, can we add this loop at other places where we are catching this exception as well?
> 
> I only see one spot where we "catch" this exception. The other areas just check and rethrow the exception. In those try blocks, they all call `ByeBufferSpanPair.captureBody` and that is where the exception is thrown.

I just checked. We should add the loop there as well. Else, we will not be blocking the requests if it's nested exception. So when we ignore the exception, the request will not be blocked.

tyroneyu:
I'm going to merge this and open a new jira to track the snyk error work. I think it's a bigger/wider ranging effort.

PR #377, opened 2023-02-24
Request body blocking for netty

shashank11p:
Adds request body blocking capability on netty


instrumentation/netty/netty-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/netty/v4_0/server/HttpServerBlockingRequestHandler.java
```diff
@@ -55,6 +57,14 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
         return;
       }
     }
+    if (msg instanceof HttpContent) {
+      if (FilterRegistry.getFilter().evaluateRequestBody(span, null, null)) {
```
ryanericson, line 61: What does this mean? It's passing null for request body?

shashank11p: We do not have the request body at this point, but we have the request body in span attributes. And the filter impl will use request body from span attributes

ryanericson (approved):

PR #379, opened 2023-04-03
Add FilterProviderConfig to pass information to FilterProvider. (Eg. service name)

tyroneyu:


filter-api/src/main/java/org/hypertrace/agent/filter/FilterRegistry.java
```diff
@@ -81,12 +83,16 @@ private static Filter load(List<String> jarPaths) {
     ClassLoader cl = loadJars(jarPaths);
     ServiceLoader<FilterProvider> providers = ServiceLoader.load(FilterProvider.class, cl);
     List<Filter> filters = new ArrayList<>();
+
+    String serviceName = HypertraceConfig.get().getServiceName();
```
tim-mwangi, line 87: Instead of this, we can modify the initialize() method to take in a FilterProviderConfig object and then at https://github.com/hypertrace/javaagent/blob/dd21236df53b29b77e80ad169b022ed5142a2bc8/otel-extensions/src/main/java/org/hypertrace/agent/otel/extensions/FilterComponentInstaller.java#L41 create it and invoke initialize with it. Right?

filter-api/src/main/java/org/hypertrace/agent/filter/FilterRegistry.java
```diff
@@ -23,8 +23,10 @@
 import java.util.Collections;
 import java.util.List;
 import java.util.ServiceLoader;
+import org.hypertrace.agent.otel.extensions.config.HypertraceConfig;
```
shashank11p, line 26: We can remove this. Then no circular dependency

tyroneyu: done. Thanks!

filter-api/src/main/java/org/hypertrace/agent/filter/FilterRegistry.java
tyroneyu: Done.

tim-mwangi (approved):
Looks good! Not sure what's with the muzzle test

shashank11p (approved):

PR #385, opened 2023-06-13
Fix netty connection keep alive

shashank11p:
When a client used `Connection: keep-alive` header, only the first request was being captured. Subsequent requests were using the same context and so we were adding data in the same first span that was ended. 

We are setting the context as null when we end the span, so that whenever a new request comes, no matter, it will create a new context and a new span. 

Tested this for both connection keep-alive and close.


instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/netty/v4_1/server/HttpServerResponseTracingHandler.java
```diff
@@ -103,6 +103,10 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise prm) {
       span.setStatus(code >= 100 && code < 500 ? StatusCode.UNSET : StatusCode.ERROR);
     }
     if (msg instanceof LastHttpContent) {
+      ctx.channel()
```
ryanericson, line 106: Sounds like this could use some comment?

shashank11p: Added

ryanericson (approved):

PR #388, opened 2023-09-25
Add content type to additional data spans

shashank11p:
In case of capturing request/response bodies in additional data spans, we only have the body. Also adding the content type header in those spans along with body.

### DIFF
--- a/instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
+++ b/instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
@@ -49,7 +49,9 @@ public class InputStreamUtils {
 
   static {
     try {
-      getAttribute = Class.forName("io.opentelemetry.sdk.trace.SdkSpan").getDeclaredMethod("getAttribute", AttributeKey.class);
+      getAttribute =
+          Class.forName("io.opentelemetry.sdk.trace.SdkSpan")
+              .getDeclaredMethod("getAttribute", AttributeKey.class);
     } catch (NoSuchMethodException e) {
       log.error("getAttribute method not found in SdkSpan class", e);
     } catch (ClassNotFoundException e) {
@@ -75,10 +77,12 @@ public class InputStreamUtils {
               .setAttribute(attributeKey, value);
 
       // Also add content type if present
-      if (getAttribute != null && span.getClass().getName().equals("io.opentelemetry.sdk.trace.SdkSpan")) {
+      if (getAttribute != null
+          && span.getClass().getName().equals("io.opentelemetry.sdk.trace.SdkSpan")) {
         try {
           Object reqContentType =
-              getAttribute.invoke(span, HypertraceSemanticAttributes.HTTP_REQUEST_HEADER_CONTENT_TYPE);
+              getAttribute.invoke(
+                  span, HypertraceSemanticAttributes.HTTP_REQUEST_HEADER_CONTENT_TYPE);
           if (reqContentType != null) {
             spanBuilder.setAttribute("http.request.header.content-type", (String) reqContentType);
           }
--- a/instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
+++ b/instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
@@ -19,6 +19,7 @@ package io.opentelemetry.javaagent.instrumentation.hypertrace.java.inputstream;
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.instrumentation.api.util.VirtualField;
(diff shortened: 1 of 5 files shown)


instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
```diff
@@ -50,12 +53,34 @@ public static void addAttribute(Span span, AttributeKey<String> attributeKey, St
...
+      SpanBuilder spanBuilder =
+          TRACER
+              .spanBuilder(HypertraceSemanticAttributes.ADDITIONAL_DATA_SPAN_NAME)
+              .setParent(Context.root().with(span))
+              .setAttribute(attributeKey, value);
+
+      // Also add content type if present
+      if (span.getClass().getName().equals("io.opentelemetry.sdk.trace.SdkSpan")) {
+        try {
+          Method getAttribute =
```
shashank11p, line 65: Needed to use reflection because this part of code will be running in the application class loader. And that will not have the sdk libraries which has this `SdkSpan` class.

instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
```diff
@@ -50,12 +53,34 @@ public static void addAttribute(Span span, AttributeKey<String> attributeKey, St
...
-      TRACER
-          .spanBuilder(HypertraceSemanticAttributes.ADDITIONAL_DATA_SPAN_NAME)
-          .setParent(Context.root().with(span))
-          .setAttribute(attributeKey, value)
-          .startSpan()
-          .end();
+      SpanBuilder spanBuilder =
+          TRACER
+              .spanBuilder(HypertraceSemanticAttributes.ADDITIONAL_DATA_SPAN_NAME)
+              .setParent(Context.root().with(span))
```
ryanericson, line 59: Why is this `Context.root()` instead of `Context.current()`?

instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
ryanericson: Can we store content type in context along with parent span? Rather than retrieving with reflection from span.

shashank11p: We can but we will need that change in all the framework instrumentations that we want to support. This was a one place change for this for all parent frameworks

instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
shashank11p: Because if the `span` has ended, then the current may not actually be the parent of this additional-data. 
It could be inside the scope of another span, but we want this additional-data with response body to be present under the `span` which we are passing down from the VirtualStore

instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
ryanericson: What method is this instrumentation on? 

Also how many framework instrumentation points do we support streaming bodies for?

shashank11p: > What method is this instrumentation on?

This is instrumented on the `read` method in `java.io.InputStream` 

> Also how many framework instrumentation points do we support streaming bodies for?

We have 4 client instrumentations and all support streaming.

instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
ryanericson: Which `span` has ended?

instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
ryanericson: Ok, my concern is with overhead, will let you take call on which approach to go. If we go with reflection, should we at least cache the Method object because that wouldn't change, will it? If that's not an expensive operation then nvm.

shashank11p: Yes, added the method to static block during instantiation. Will be called only once.

instrumentation/java-streams/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/java/inputstream/InputStreamUtils.java
shashank11p: So if we had a root span -> parent span (which has the inputstream).
Now, the parent span is ended, but the inputstream.read() is not yet called, but the inputstream is passed down, and another span is created, and then the input stream is read. Then the current span where input stream is read is not the actual span under which we want to create the additional data span. So that may the reason. We would want it like:
```
root span:
  parent span(ended):
    additional span:
  new span(calls inputstream.read()):
```
I wasn't involved when this code was written, so I'm also trying to think and justify this.

ryanericson (approved):

PR #398, opened 2023-12-14
Custom block code and msg

shashank11p:
Adding a FilterResult in the Filter API, so that filters can provide custom blocking status code and message along with the FilterResult. 
Using this FilterResult to block and set the response status and message accordingly

### DIFF
--- a/filter-api/src/main/java/org/hypertrace/agent/filter/MultiFilter.java
+++ b/filter-api/src/main/java/org/hypertrace/agent/filter/MultiFilter.java
@@ -19,6 +19,7 @@ package org.hypertrace.agent.filter;
 import io.opentelemetry.api.trace.Span;
 import java.util.List;
 import java.util.Map;
+import org.hypertrace.agent.core.filter.FilterResult;
 import org.hypertrace.agent.filter.api.Filter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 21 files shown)


filter-api/src/main/java/org/hypertrace/agent/filter/MultiFilter.java
```diff
@@ -48,16 +49,16 @@ public boolean evaluateRequestHeaders(Span span, Map<String, String> headers) {
             t);
       }
     }
-    return shouldBlock;
+    return new FilterResult(false, 403, "Hypertrace Blocked Request");
```
varkey98, line 52: the default message and status code would be 200 and OK right? since we 're setting shouldBlock to false

varkey98: yes its a nit only, I get that we would be overriding this in our filter implementation and incase of pass through it wont be set at all, but just for consistency

shashank11p: Updated to empty msg and `0` as status code since they should not be used

varkey98 (approved):

ryanericson (commented):
I suppose this is pending discussion for GRPC

shashank11p:
> I suppose this is pending discussion for GRPC

Yes, for now it's setting the Status description as the msg provided. I'll just revert that in grpc, and we will update grpc when we finalise how to set it for grpc.

shashank11p:
> > I suppose this is pending discussion for GRPC
> 
> Yes, for now it's setting the Status description as the msg provided. I'll just revert that in grpc, and we will update grpc when we finalise how to set it for grpc.

Reverted these. Right now, we do not add custom status and msgs for grpc. 
We have added them for all other frameworks that we support

PR #399, opened 2024-01-19
add grpc mappings for custom blocking status code

shashank11p:


instrumentation/grpc-1.6/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/grpc/v1_6/server/GrpcServerInterceptor.java
```diff
@@ -78,6 +78,32 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
...
+        return Status.UNAUTHENTICATED;
+      case 403:
+        return Status.PERMISSION_DENIED;
+      case 404:
+        return Status.UNIMPLEMENTED;
+      case 429:
+        return Status.UNAVAILABLE;
+      case 502:
+        return Status.UNAVAILABLE;
+      case 503:
```
varkey98, line 98: nit: should 5xx be mapped to INTERNAL ?

shashank11p: No, We are using https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md as specified

varkey98 (approved):

PR #400, opened 2024-01-23
update otel to 1.32.0

kishansairam9:


build.gradle.kts
```diff
@@ -54,14 +54,11 @@ subprojects {
     extra.set("versions", mapOf(
         // when updating these values, some values must also be updated in buildSrc as this map
         // cannot be accessed there
-        "opentelemetry" to "1.24.0",
-        "opentelemetry_semconv" to "1.24.0-alpha",
```
kishansairam9, line 58: `io.opentelemetry:opentelemetry-semconv` was used before which wasn't present in bom of opentelmetry instrumentation, but `io.opentelemetry.semconv:opentelemetry-semconv` was present.
https://repo.maven.apache.org/maven2/io/opentelemetry/instrumentation/opentelemetry-instrumentation-bom-alpha/1.32.0-alpha/opentelemetry-instrumentation-bom-alpha-1.32.0-alpha.pom
```
<dependency>
<groupId>io.opentelemetry.semconv</groupId>
<artifactId>opentelemetry-semconv</artifactId>
<version>1.21.0-alpha</version>
</dependency>
```

kishansairam9: https://github.com/open-telemetry/opentelemetry-java/blob/e4b31cea3de13db7262540457e598282f56d9978/CHANGELOG.md?plain=1#L231

they stopped publishing `io.opentelemetry:opentelemetry-semconv` and suggest to use the new one

PR #406, opened 2024-08-19
feat: remove high cardinality metrics

thugrock7:
## Description
Do not record high cardinality attributes in metrics which could potentially cause max cardinality errors and lead to drop of all attributes


shashank11p (commented):
Please add some tests for metrics as well. Maybe in smoke tests. We already have checks for some metrics. We can check for attributes in metrics there as well

shashank11p (approved):

PR #411, opened 2024-12-17
feat: servlet 5.0 body capture fix

thugrock7:


instrumentation/servlet/servlet-5.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/servlet/v5_0/Servlet50AndFilterInstrumentation.java
```diff
@@ -187,6 +187,7 @@ public static void exit(
 
         if (!request.isAsyncStarted()) {
           if (instrumentationConfig.httpHeaders().response()) {
+            httpResponse.flushBuffer();
```
shashank11p, line 190: why do we need this?

thugrock7: The HttpServletResponse `getHeaderNames()` will only give headers that are set via set/add Headers functions at client side ([docs](https://jakarta.ee/specifications/servlet/5.0/apidocs/jakarta/servlet/http/httpservletresponse#getHeaderNames())), so for servlet applications where content-type is set in general with [setContentType()](https://jakarta.ee/specifications/servlet/5.0/apidocs/jakarta/servlet/servletresponse#setContentType(java.lang.String)), at this specific point if we call `getHeaderNames()`, it will not contain content-type, content-length etc headers as it is explicitly not set by setHeader()/addHeader() at client side. So we will not collect the content-type header and thereby dropping the body. (For context look at the simple application [here](https://github.com/Traceableai/qa-test-applications/blob/main/http-applications/java/http-servlet/src/main/java/com/traceableai/servlet/App.java#L164))

The info which is set by setContentType(), setContentEncoding() etc are available in buffer, which will be set as headers to client at close commit normally, so calling `flushbuffer()` ourselves on OnMethodExit() before to get these headers populated into span.

shashank11p (approved):

PR #414, opened 2025-02-07
feat: use shaded protobuf-java , instead of applications

thugrock7:


instrumentation/grpc-1.6/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/grpc/v1_6/GrpcSpanDecorator.java
```diff
@@ -44,12 +41,13 @@ public static void addMessageAttribute(Object message, Span span, AttributeKey<S
     if (message instanceof Message) {
       Message mb = (Message) message;
       try {
-        BoundedCharArrayWriter writer = BoundedBuffersFactory.createWriter();
-        PRINTER.appendTo(mb, writer);
-        span.setAttribute(key, writer.toString());
-      } catch (IOException e) {
-        log.error("Failed to decode message to JSON", e);
+        String jsonOutput = ProtobufMessageConverter.getMessage(mb);
```
shashank11p, line 44: Why do we need to do this? What was happening while we were directly trying to get it?

PR #419, opened 2025-04-22
feat: fix support of transitive dependencies in conversion of protobuf messages to json

thugrock7:


instrumentation/grpc-1.6/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/grpc/v1_6/ProtobufMessageConverter.java
```diff
@@ -106,36 +157,29 @@ private static Descriptor getRelocatedDescriptor(Descriptors.Descriptor original
...
+    return relocatedFileDescriptor;
   }
 
   /**
    * Method that takes an incoming message, converts it to a relocated one, prints it as JSON using
    * the relocated JsonFormat
    *
    * @param message The incoming (unrelocated) protobuf message.
+   * @return JSON string representation of the message
+   * @throws Exception if conversion fails
```
shashank11p, line 169: all usages of this function have try catch to handle this exception?

shashank11p (approved):

instrumentation/grpc-1.6/src/main/java/io/opentelemetry/javaagent/instrumentation/hypertrace/grpc/v1_6/ProtobufMessageConverter.java
thugrock7: yes, there is single usage for this function ([here](https://github.com/hypertrace/javaagent/pull/419/files#diff-14f97641c3a25445cb8662900a3e16116956c8ad80d501f76233201859cf7284R43)), and that wrapper is used at multiple places. So we are handling this safely

PR #420, opened 2025-04-22
feat: using latest ubuntu runners

thugrock7:
Description: 
Ubuntu 20.04 Actions runner image is deprecated as of 15th April 2025. So updating to use ubuntu-latest image
Ref: https://github.com/actions/runner-images/issues/11101
SInce latest ubuntu image uses java 17 by default, updated to use JDK 11 to build and test to keep things as it is

### DIFF
--- a/.github/workflows/build-springboot-smoke-dist.yml
+++ b/.github/workflows/build-springboot-smoke-dist.yml
@@ -42,3 +42,5 @@ jobs:
           ./gradlew jib -PtargetJDK=11 -Djib.httpTimeout=120000 -Djib.console=plain --stacktrace -Ptag=$TAG
           ./gradlew jib -PtargetJDK=15 -Djib.httpTimeout=120000 -Djib.console=plain --stacktrace -Ptag=$TAG
         working-directory: smoke-tests/springboot
+        env:
+          JVM_OPTS: -Xmx1g --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED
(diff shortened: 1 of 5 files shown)


.github/workflows/build-springboot-smoke-dist.yml
```diff
@@ -42,3 +42,5 @@ jobs:
           ./gradlew jib -PtargetJDK=11 -Djib.httpTimeout=120000 -Djib.console=plain --stacktrace -Ptag=$TAG
           ./gradlew jib -PtargetJDK=15 -Djib.httpTimeout=120000 -Djib.console=plain --stacktrace -Ptag=$TAG
         working-directory: smoke-tests/springboot
+        env:
+          JVM_OPTS: -Xmx1g --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED
```
varkey98, line 46: what are these java options for?

varkey98 (approved):

.github/workflows/build-springboot-smoke-dist.yml
thugrock7: The `--add-opens` flags were added because Java 17 (default in ubuntu-latest) strictly blocks reflection access that was permitted in Java 11.
Ref actions: https://github.com/hypertrace/javaagent/actions/runs/14587461988/job/40915439308
To solve `InaccessibleObjectException` observed, explicitly granting permission to use reflection on Java's internal classes of `java.util` and `java.io` packages

shashank11p (approved):

PR #433, opened 2026-07-21
chore: Bump opentelemetry jackson version

thugrock7:
Upgrading vulnerable `jackson-databind` dependency coming from otel javaagent JAR

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -64,7 +64,8 @@ subprojects {
 
         "opentelemetry_gradle_plugin" to "1.33.0-alpha",
         "byte_buddy" to "1.12.10",
-        "slf4j" to "2.0.7"
+        "slf4j" to "2.0.7",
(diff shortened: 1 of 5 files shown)


tim-mwangi (approved):

build.gradle.kts
```diff
@@ -64,7 +64,8 @@ subprojects {
 
         "opentelemetry_gradle_plugin" to "1.33.0-alpha",
         "byte_buddy" to "1.12.10",
-        "slf4j" to "2.0.7"
+        "slf4j" to "2.0.7",
+        "jackson" to "2.18.9"
```
tim-mwangi, line 68: This is the same version as before?

thugrock7: jackson bumped from 2.16.1 to 2.18.9
the `"slf4j" to "2.0.7",` is being shown because it has `,` added in it

